PR #5, opened 2020-11-09
add execute-in-vm and cleanup-vm task

suomiy:
Adds two additional tasks:

`execute-in-vm` task  can execute a script, or a command in a Virtual Machine over SSH. The execution of a command can be implemented by other backends in the future. [README](https://github.com/suomiy/kubevirt-tekton-tasks/blob/execute-in-vm/tasks/execute-in-vm/README.md)

`cleanup-vm` task  can execute a script, or a command in a Virtual Machine and stop/delete 
the VM afterwards. It is best used together with tekton pipelines finally construct. It uses execute-in-vm image as an implementation. [README](https://github.com/suomiy/kubevirt-tekton-tasks/blob/execute-in-vm/tasks/cleanup-vm/README.md)

Todo:
- [x] unit tests
- [x] functional tests


suomiy:
added unit tests

PR #49, opened 2021-02-12
README: improve English style, add motivation

dankenigsberg:
I try to explain why running your tests inside a VM may be required.


suomiy:
Thanks for the improvements

PR #60, opened 2021-03-01
add disk-virt-sysprep task

suomiy:
Introduces a task which will run `virt-sysprep` on a target pvc: [README](https://github.com/suomiy/kubevirt-tekton-tasks/tree/virt-sysprep/tasks/disk-virt-sysprep). It should be similar to [virt-customize](https://github.com/kubevirt/kubevirt-tekton-tasks/pull/34) task.

cc @fabiand @jelkosz 

@fzdarsky can you please check the API if it covers all your use cases? 

1. Do you need support for attaching multiple PVCs?
2. Do you need support for uploading some files/k8s objects to the VM's disk by using `--upload` command?


fzdarsky:
@suomiy sorry for the slow response.

> @fzdarsky can you please check the API if it covers all your use cases? 
LGTM.

> 1. Do you need support for attaching multiple PVCs?
No, single PVC for rootdisk is ok for us.

> 2. Do you need support for uploading some files/k8s objects to the VM's disk by using `--upload` command?
I assume the `virt-customize` task will support that, or? In that case, no, we wouldn't need it for `virt-sysprep` task, too.

suomiy:
> > 1. Do you need support for uploading some files/k8s objects to the VM's disk by using `--upload` command?
> 
> I assume the `virt-customize` task will support that, or? In that case, no, we wouldn't need it for `virt-sysprep` task, too.

Both `virt-customize` and `virt-sysprep` will support it in a sense that the library is able to conduct these commands. But it will not be useful since the only availbale source destinations would be in the task container itself. 
For it  to be useful we would have to come up with some input; like mounting ConfigMap / PVC / etc. to the container.

PR #64, opened 2021-03-05
distribute only namespace scoped RBAC + add README

suomiy:
I tried to simplify the tasks installation by removing cluster/namespace rbac commands from tasks' READMEs. And by introducing  tasks-rbac-permissions README which should describe all other permission use cases.  

This PR will be followed up by https://github.com/kubevirt/kubevirt-tekton-tasks/pull/51 which will remove Install section from tasks' READMEs and introduce a single installation process in the main README. 

@dankenigsberg Can you please take a look? ([tasks-rbac-permissions.md](https://github.com/suomiy/kubevirt-tekton-tasks/blob/rbacUpdate/docs/tasks-rbac-permissions.md) should be enough)


docs/tasks-rbac-permissions.md
```diff
@@ -0,0 +1,140 @@
...
+Please see Tekton documentation on how to specify service accounts in
+[TaskRuns](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#configuring-a-taskrun)
+and [PipelineRuns](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#mapping-serviceaccount-credentials-to-tasks).
+
+This will make sure that the task's pod has right permissions for accessing the resources it needs.
+For example `execute-in-vm` task has RBAC permissions for fetching `VirtualMachines` and for executing `start`/`stop` operations on them.
+
+
+### Example service account mapping
+The name of the service account is usually named `${TASK_NAME}-task`.
```
dankenigsberg, line 19:
```suggestion
The service account is usually named `${TASK_NAME}-task`.
```

docs/tasks-rbac-permissions.md
```diff
@@ -0,0 +1,140 @@
...
+done
+```
+
+This example enables `generate-ssh-keys` and `create-vm-from-manifest` tasks to create secrets/VMs in any namespace in the cluster from a pipeline started in `ns-a`.
+
+Warning: this will allow users with access to `generate-ssh-keys-task` and `create-vm-from-manifest-task` service accounts to run pods in any namespace.
+This will elevate their privileges to cluster admin level.
+
+
+# Running the tasks in new namespaces
```
dankenigsberg, line 107: is "running" as task different from "deploying" it? If not, can you use the same verb? it seems that `new-ns` behaves like `ns-a`, but I'm not sure.

Maybe you can use `task-ns1` and `vm-ns1` and then add  `task-ns2` and `vm-ns2` so it's clearer for a layperson?

dankenigsberg (changes requested):

docs/tasks-rbac-permissions.md
```diff
@@ -0,0 +1,140 @@
+# RBAC permissions for running the tasks
+
+Each KubeVirt Tekton Task usually comes with its own `ServiceAccount`, `ClusterRole` and `RoleBinding`.
+This allows the task to monitor, create, update and delete relevant resources
+in the namespace where KubeVirt Tekton Tasks were deployed.
```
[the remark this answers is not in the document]
suomiy: That is much better, thanks.

docs/tasks-rbac-permissions.md
suomiy: fixed

docs/tasks-rbac-permissions.md
suomiy: That is indeed clearer. Changed it to use these namespace. Although `vm-ns2` was not necessary.

dankenigsberg: But it is necessary to me: I'm trying to understand how one piplinerun can access resources from two different namespaces

suomiy: Hm, in that case I could add 2nd loop for the namespace to https://github.com/suomiy/kubevirt-tekton-tasks/blob/rbacUpdate/docs/tasks-rbac-permissions.md#multi-namespace-rbac-permissions example.

Although that might be too much for a simple use case regarding the readability?

dankenigsberg: I would drop `task-ns2` for brevity. It's somehow more obvious to me to me that I can deploy the task wherever I want.

suomiy: - I added two namespaces to [#multi-namespace-rbac-permissions](https://github.com/suomiy/kubevirt-tekton-tasks/blob/rbacUpdate/docs/tasks-rbac-permissions.md#multi-namespace-rbac-permissions) example. Please let me know if the readability suffers.
- I think It is good to keep the `task-ns2` because there has to some namespace specified in [#deploying-the-tasks-in-additional-namespaces](https://github.com/suomiy/kubevirt-tekton-tasks/blob/rbacUpdate/docs/tasks-rbac-permissions.md#deploying-the-tasks-in-additional-namespaces) section.

suomiy: > I think It is good to keep the task-ns2 because there has to be some namespace specified in #deploying-the-tasks-in-additional-namespaces section.

My wrong, I guess there doesn't have to be - I suppose we could remove it. 

Not sure what is better though - verbosity(more clear) or reusability.

suomiy:
merging because of https://github.com/kubevirt/kubevirt-tekton-tasks/pull/64#discussion_r590371997

docs/tasks-rbac-permissions.md
dankenigsberg: Sorry, I've lost you here. Could you point me to the current description of multi-namespace usage, or that follow-up PR?

PR #80, opened 2021-04-01
add windows-installer example pipeline

suomiy:
Downloads a Windows Source ISO into a PVC and installs Windows into a new root PVC.
It uses Windows Sysprep to automatically install the system. 
The pipeline can be customized to support different installation requirements. 

The artifact of this pipeline is a PVC (backed by DataVolume) with a custom Windows installation. 

[README](https://github.com/suomiy/kubevirt-tekton-tasks/tree/win-pipeline/examples/pipelines/windows-installer)

Todo:
- [x] implement wait-for-vm task (https://github.com/kubevirt/kubevirt-tekton-tasks/pull/81)

### DIFF
--- a/examples/pipelines/windows-installer/README.md
+++ b/examples/pipelines/windows-installer/README.md
@@ -1,0 +1,85 @@
+# Windows Installer Pipeline
+
+Downloads a Windows Source ISO into a PVC and installs Windows into a new base PVC.
+It will then spin up an installation VM and use Windows Answer Files to automatically install the VM.
+Then the pipeline will wait for the installation to complete and will delete the installation VM while keeping the artifact PVC (backed by DataVolume) with the installed operating system.
+The pipeline can be customized to support different installation requirements.
+
+## Prerequisites
+
+- KubeVirt `v0.39.0`
+- Tekton Pipelines `v0.19.0`
+
+### Obtain Windows ISO download URL
+
+1. Go to https://www.microsoft.com/en-us/software-download/windows10ISO. You can also obtain a server edition for evaluation at https://www.microsoft.com/en-us/evalcenter/evaluate-windows-server-2019.
+2. Fill in the edition and `English` language (other languages need to be updated in windows-10-autounattend ConfigMap) and go to the download page.
+3. Right-click on the 64-bit download button and copy the download link. The link should be valid for 24 hours.
+4. Initialize a WIN_URL variable that will be used to create a DataVolume which will download this ISO into a PVC.
+
+```bash
+# Real URL can look differently
+WIN_URL="https://software-download.microsoft.com/db/Win10_20H2_v2_English_x64.iso..."
+```
+
+### Prepare autounattend.xml ConfigMap
+
+1. Supply, generate or use the default autounattend.xml.
+   The configuration file can be generated with [Windows SIM](https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/wsim/windows-system-image-manager-overview-topics)
+   or it can be specified manually according to [Answer File Reference](https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/wsim/answer-files-overview)
+   and [Answer File Components Reference](https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/components-b-unattend).
+2. Replace the default example autounattend.xml with your own in `windows-installer-pipeline.yaml`.
+   You can also store the config inside a secret (VM definition requires changes in that case).
+   Different autounattend.xml can be also passed in the Pipeline parameters when creating a PipelineRun.
+
+## Pipeline Description
+
+```
+  create-source-dv --- create-vm-from-manifest --- wait-for-vmi-status --- cleanup-vm
+                    |
+    create-base-dv --
+```
+
+1. `create-source-dv` task downloads a Windows source ISO into a PVC called `windows-10-source-*`.
+2. `create-base-dv` task creates empty PVC for new windows installation called `windows-10-base-*`.
+3. `create-vm-from-manifest` task creates a VM called `windows-installer-*`
+   from the empty PVC and with `windows-10-source-*` PVC attached as a CD-ROM.
+4. ` wait-for-vmi-status` task waits until the VM shutdowns.
+5. `cleanup-vm` deletes the installer VM and ISO PVC.
+6.  The output artifact will be the `windows-10-base-*` PVC with the Windows installation. 
+    It includes an `Administrator` user with `changepassword` password.
+
+## How to run
+
+```bash
+WIN_URL="https://software-download.microsoft.com/db/Win10_20H2_v2_English_x64.iso..."
+kubectl apply -f windows-installer-pipeline.yaml
+sed 's!DOWNLOAD_URL!'"$WIN_URL"'!g' windows-installer-pipelinerun.yaml | kubectl create -f -
+```
+
+## Possible Optimizations
+
+### Windows Source ISO Caching
+
+Windows source ISO can be downloaded and cached/stored in a PVC.
+So the subsequent PipelineRuns don't have to download the same ISO multiple times.
+
+You can create the following DV first and remove create-source-dv step from the windows-installer-pipeline:
+
+```yaml
+ apiVersion: cdi.kubevirt.io/v1beta1
+kind: DataVolume
+metadata:
+  name: windows-10-source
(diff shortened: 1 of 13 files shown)


suomiy:
@dankenigsberg @fabiand

I decided to split the Windows flow into 2 pipelines.

1. This one which installs the windows and has a PVC as an artifact.
2. Pipeline which copies this base PVC and runs a build/deployment from that. 

IMO a single pipeline with these two use cases would be too cumbersome.

Also this pipeline requires a `wait-for-vm` task which will detect that the VM turned off and the installation has completed. 
This task needs to be implemented before this pipeline. 

Any suggestions for a better name than `wait-for-vm`?

examples/pipelines/windows-installer/README.md
```diff
@@ -0,0 +1,60 @@
...
+```bash
+WIN_URL="https://software-download.microsoft.com/db/Win10_20H2_v2_English_x64.iso..."
+kubectl apply -f windows-installer-pipeline.yaml
+sed 's!DOWNLOAD_URL!'"$WIN_URL"'!g' windows-installer-pipelinerun.yaml | kubectl create -f -
+```
+
+## Possible Optimizations
+
+- Windows source ISO can be downloaded and cached/stored in a PVC. 
+  So the subsequent PipelineRuns don't have to download the same ISO multiple times.
```
dankenigsberg, line 60: Can you provide the optimization right now? The pipeline should better work without Internet connectivity.

examples/pipelines/windows-installer/README.md
```diff
@@ -0,0 +1,60 @@
...
+4. Initialize a WIN_URL variable that will be used to create a DataVolume which will download this ISO into a PVC.
+
+```bash
+# Real URL can look differently
+WIN_URL="https://software-download.microsoft.com/db/Win10_20H2_v2_English_x64.iso..."
+```
+
+### Prepare autounattend.xml ConfigMap
+
+1. Supply, generate or use the default autounattend.xml.
```
dankenigsberg, line 28: I'm not sure how I can generate or where I can supply my own autounattend.xml file. I suppose that a usage example of the suggested Pipeline would help.

dankenigsberg (changes requested):
> I decided to split the Windows flow into 2 pipelines.
> 
>     1. This one which installs the windows and has a PVC as an artifact.

Ok, as this first pipeline seems useful on its own right. I'd love to tweak it so it places its artifact under openshift-virtualization-os-images namespace to be auto-cloned by GUI when a new VM is requested.

>     2. Pipeline which copies this base PVC and runs a build/deployment from that.

Not sure what this consists.
> 
> Also this pipeline requires a `wait-for-vm` task which will detect that the VM turned off and the installation has completed.
> This task needs to be implemented before this pipeline.
> 
> Any suggestions for a better name than `wait-for-vm`?

Should this task wait for a vm to shut down? If so, I'd call it `wait-for-vm-shutdown`. Does Tekton have similar tasks for waiting for pods?

examples/pipelines/windows-installer/windows-installer-pipeline.yaml
```diff
@@ -0,0 +1,320 @@
...
+                    <SynchronousCommand wcm:action="add">
+                        <Order>3</Order>
+                        <RequiresUserInput>false</RequiresUserInput>
+                        <CommandLine>cmd /C wmic useraccount where name="Administrator" set PasswordExpires=false</CommandLine>
+                        <Description>Password Never Expires</Description>
+                    </SynchronousCommand>
+                    <SynchronousCommand wcm:action="add">
+                        <Order>4</Order>
+                        <RequiresUserInput>false</RequiresUserInput>
+                        <CommandLine>cmd /c shutdown /s /f /t 10</CommandLine>
```
eifrach, line 178: I'm missing here the sysprep command.
you mention it in the readme  but no reference to it

suomiy:
> > I decided to split the Windows flow into 2 pipelines.
> > ```
> > 1. This one which installs the windows and has a PVC as an artifact.
> > ```
> 
> Ok, as this first pipeline seems useful on its own right. I'd love to tweak it so it places its artifact under openshift-virtualization-os-images namespace to be auto-cloned by GUI when a new VM is requested.

Yes, it can be used this way as one of the use case , but we should  keep it runnable on Kubernetes in upstream as well.

> 
> > ```
> > 2. Pipeline which copies this base PVC and runs a build/deployment from that.
> > ```
> 
> Not sure what this consists.

this is TBD

> 
> > Also this pipeline requires a `wait-for-vm` task which will detect that the VM turned off and the installation has completed.
> > This task needs to be implemented before this pipeline.
> > Any suggestions for a better name than `wait-for-vm`?
> 
> Should this task wait for a vm to shut down? If so, I'd call it `wait-for-vm-shutdown`. Does Tekton have similar tasks for waiting for pods?

They have a similar thing for [kubectl-deploy-pod](https://github.com/tektoncd/catalog/tree/main/task/kubectl-deploy-pod/0.1), but I would rather keep it in a separate task since VM's have more complex lifecycle.

They use `success-condition` and `failure-condition`, so we could do something similar to that. I would prefer to have a single task rather than multiple of `wait-for-vm-shutdown`, etc. Since you might want to wait for Running / Succeeded / Failed

examples/pipelines/windows-installer/windows-installer-pipeline.yaml
suomiy: It should be run automatically by the windows installer when it locates `autounattend.xml`

examples/pipelines/windows-installer/README.md
suomiy: I can put a link to the Microsoft documenation there if that helps.

examples/pipelines/windows-installer/README.md
suomiy: I would prefer for the pipeline to be as simple as possible so one can run the example quickly.  But I can describe the extra steps here in this chapter.

examples/pipelines/windows-installer/README.md
dankenigsberg: A link would help, but I'm asking for an example of a pipeline with a non-default `autounattended.xml`

dankenigsberg:
> They use success-condition and failure-condition, so we could do something similar to that. I would prefer to have a single task rather than multiple of wait-for-vm-shutdown, etc. Since you might want to wait for Running / Succeeded / Failed

`wait-for-vm-status` then?

examples/pipelines/windows-installer/README.md
suomiy: I am planning to add ssh server installation to this pipeline later. But other use case specific flow can be implemented as an addon / other pipeline later.

suomiy:
> > They use success-condition and failure-condition, so we could do something similar to that. I would prefer to have a single task rather than multiple of wait-for-vm-shutdown, etc. Since you might want to wait for Running / Succeeded / Failed
> 
> `wait-for-vm-status` then?

1. Shouldn't we rather name it `wait-for-vmi-status` since it will be checking VMI status? 
2. It might include checking for VM status in the future, but that is probably too much of an edge case

dankenigsberg:
> 
>     1. Shouldn't we rather name it `wait-for-vmi-status` since it will be checking VMI status?

wfm

> 
>     2. It might include checking for VM status in the future, but that is probably too much of an edge case

I don't understand what you're saying here

suomiy:
> > ```
> > 1. Shouldn't we rather name it `wait-for-vmi-status` since it will be checking VMI status?
> > ```
> 
> wfm
> 
> > ```
> > 2. It might include checking for VM status in the future, but that is probably too much of an edge case
> > ```
> 
> I don't understand what you're saying here

if it should have other detection abilities in the future with regards to this status: https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachinestatus

suomiy:
PR for wait-for-vmi-status task: #81

examples/pipelines/windows-installer/README.md
suomiy: fixed

examples/pipelines/windows-installer/README.md
suomiy: fixed

suomiy:
#81 has been merged and the pipeline should be getting ready for merging

suomiy:
merging for https://github.com/kubevirt/kubevirt.github.io/pull/727

PR #93, opened 2021-08-02
add copy-template task

ksimon1:
this task copies template which exists in cluster and
creates a new one with different name and namespace.

**Release note**:
```
Add copy template task
```


modules/copy-template/cmd/copy-template/main.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	log.Logger().Debug("parsed arguments", zap.Reflect("cliOptions", cliOptions))
+
+	templateCreator, err := templatecreator.NewTemplateCreator(cliOptions)
+	if err != nil {
+		exit.ExitOrDieFromError(GenericExitCode, err)
+	}
+
+	newTemplate, err := templateCreator.CreateTemplateFromTemplate()
```
kwiesmueller, line 39: Could this be renamed to `CopyFromTemplate` or `CopyTemplate` or is there a difference between the two?

modules/copy-template/pkg/templates/template-provider.go
```diff
@@ -0,0 +1,79 @@
...
+	updatedTemplate, err := t.UpdateTemplate(template)
+	if err != nil {
+		return nil, err
+	}
+
+	return t.templateProvider.Create(t.cliOptions.GetTargetTemplateNamespace(), updatedTemplate)
+}
+
+func (t *TemplateCreator) UpdateTemplate(template *v1.Template) (*v1.Template, error) {
+	template.ResourceVersion = ""
```
kwiesmueller, line 75: Is that the only field that needs resetting?
I could also think of `uuid`, `managedFields` and a few others.
To be safe you could also create a new `ObjectMeta` for the new template and only take over Name, Namespace, Labels and Annotations.

modules/copy-template/pkg/utils/parse/clioptions.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type CLIOptions struct {
+	SourceTemplateName      string `arg:"--source-template-name,env:SOURCE_TEMPLATE_NAME" placeholder:"NAME" help:"Name of a source template"`
+	SourceTemplateNamespace string `arg:"--source-template-namespace,env:SOURCE_TEMPLATE_NAMESPACE" placeholder:"NAMESPACE" help:"Namespace of a source template"`
+	TargetTemplateName      string `arg:"--target-template-name,env:TARGET_TEMPLATE_NAME" placeholder:"NAME" help:"Name of a target template"`
+	TargetTemplateNamespace string `arg:"--target-template-namespace,env:TARGET_TEMPLATE_NAMESPACE" placeholder:"NAMESPACE" help:"Namespace of a target template"`
+
+	Output output.OutputType `arg:"-o" placeholder:"FORMAT" help:"Output format. One of: yaml|json"`
+	Debug  bool              `arg:"--debug" help:"Sets DEBUG log level"`
```
kwiesmueller, line 25: Does Debug not have an env variable on purpose?

templates/copy-template/readmes/README.md
```diff
@@ -0,0 +1,29 @@
+# Copy okd template Task
+
+This task copies template.
```
kwiesmueller, line 3:
```suggestion
This task copies a template.
```

kwiesmueller (commented):
Looks good, just some minor nits and questions.

templates/copy-template/manifests/copy-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+spec:
+  params:
+    - name: sourceTemplateName
+      description: Name of an OKD template to copy template from.
+      type: string
+    - name: sourceTemplateNamespace
+      description: Namespace of an source OKD template to copy template from. (defaults to active namespace)
+      type: string
+    - name: targetTemplateName
+      description: Name of an target OKD template.
```
atiratree, line 30: we can generate the name of the targetTemplate if not specified

templates/copy-template/readmes/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Service Account
+
+This task should be run with `{{task_yaml.metadata.annotations['task.kubevirt.io/associatedServiceAccount']}}` serviceAccount.
+Please see [RBAC permissions for running the tasks](../../docs/tasks-rbac-permissions.md) for more details.
+
+### Parameters
+
+{% for item in task_yaml.spec.params %}
+{% if item.name == "templateParams" %}
```
atiratree, line 14: we don't have templateParams here

templates/copy-template/readmes/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Results
+
+{% for item in task_yaml.spec.results %}
+- **{{ item.name }}**: {{ item.description | replace('"', '`') }}
+{% endfor %}
+
+### Usage
+
+Please see [examples](examples) on how to do a copy template from a template.
```
atiratree, line 29: maybe just: `on how to copy a template?`

atiratree (changes requested):

modules/tests/test/copy_template_test.go
```diff
@@ -0,0 +1,168 @@
...
+					TargetTemplateNamespace: openshiftNamespace,
+				},
+			}),
+		)
+	})
+	Context("copy template sucess", func() {
+		table.DescribeTable("taskrun succeded and template is created", func(config *testconfigs.CopyTemplateTestConfig, targetNamespace string) {
+			f.TestSetup(config)
+			if template := config.TaskData.Template; template != nil {
+				f.TemplateClient.Templates(template.Namespace).Create(context.TODO(), template, v1.CreateOptions{})
```
atiratree, line 122: the template needs to be managed - for cleanup

modules/tests/test/testconfigs/copy-template-config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type CopyTemplateTestConfig struct {
+	TaskRunTestConfig
+	TaskData CopyTemplateTaskData
+
+	deploymentNamespace string
+}
+
+func (c *CopyTemplateTestConfig) Init(options *testoptions.TestOptions) {
```
atiratree, line 27: please take a look at other tasks on how to do a init function with namespaces

atiratree (changes requested):
1. review concerns YAMLs
2. review concerns code

templates/copy-template/manifests/copy-template.yaml
```diff
@@ -0,0 +1,63 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: ClusterTask
+metadata:
+  annotations:
+    task.kubevirt.io/associatedServiceAccount: {{ sa_name }}
+    sourceTemplateName.params.task.kubevirt.io/type: {{ task_param_types.vm_template_name }}
```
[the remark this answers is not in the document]
ksimon1: what default string you mean?

templates/copy-template/manifests/copy-template.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - copy-template
+      args:
+        - "--output=yaml"
+        - '--source-template-name'
+        - $(params.sourceTemplateName)
+        - '--source-template-namespace'
+        - $(params.sourceTemplateNamespace)
+        - '--target-template-name'
+        - $(params.targetTemplateName)
+        - '--target-template-namespace'
```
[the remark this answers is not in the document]
ksimon1: What is passed twice?

templates/copy-template/manifests/copy-template.yaml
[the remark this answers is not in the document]
atiratree: - the type does not have to be VM template - but can be any type of template
- the types are only used here if they differ from the original one which is string (https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters)

modules/create-vm/pkg/vmcreator/vm-creator.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/kubevirt/kubevirt-tekton-tasks/modules/create-vm/pkg/pvc"
 	"github.com/kubevirt/kubevirt-tekton-tasks/modules/create-vm/pkg/templates"
 	"github.com/kubevirt/kubevirt-tekton-tasks/modules/create-vm/pkg/templates/validations"
-	"github.com/kubevirt/kubevirt-tekton-tasks/modules/create-vm/pkg/utils/log"
+	"github.com/kubevirt/kubevirt-tekton-tasks/modules/shared/pkg/log"
```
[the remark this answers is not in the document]
ksimon1: Done https://github.com/kubevirt/kubevirt-tekton-tasks/pull/98

ksimon1:
/hold
This PR https://github.com/kubevirt/kubevirt-tekton-tasks/pull/98 has to be merged before this PR

modules/tests/test/copy_template_test.go
```diff
@@ -43,95 +37,59 @@ var _ = Describe("Copy template task", func() {
...
 					LimitTestScope: NamespaceTestScope,
 				},
-
 				TaskData: testconfigs.CopyTemplateTaskData{
-					Template:                cirrosTemplate,
-					SourceTemplateName:      cirrosTemplate.Name,
-					SourceTemplateNamespace: cirrosTemplate.Namespace,
-					TargetTemplateName:      newTemplateName,
-					TargetTemplateNamespace: openshiftNamespace,
+					ForbiddenNamespace: openshiftNamespace,
```
atiratree, line 64: Is havingn additional variable necessary? Can we just use SystemTargetNS ?

modules/tests/test/copy_template_test.go
```diff
@@ -141,28 +99,49 @@ var _ = Describe("Copy template task", func() {
...
 				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
 					ServiceAccount: CopyTemplateServiceAccountName,
 					LimitTestScope: ClusterTestScope,
 				},
 				TaskData: testconfigs.CopyTemplateTaskData{
-					Template:                cirrosTemplate,
-					SourceTemplateName:      cirrosTemplate.Name,
-					SourceTemplateNamespace: cirrosTemplate.Namespace,
-					TargetTemplateName:      newTemplateName,
 					TargetTemplateNamespace: openshiftNamespace,
```
atiratree, line 110: We should not polute other namespaces with test resources. There is a deploy DeployTargetNS a TestTargetNS for this use case.

modules/tests/test/testconfigs/copy-template-config.go
```diff
@@ -25,7 +32,28 @@ type CopyTemplateTestConfig struct {
...
+
+	c.TaskData.SourceNamespace = options.ResolveNamespace(c.TaskData.SourceTemplateNamespace, options.TestNamespace)
+
+	if c.TaskData.ForbiddenNamespace != "" {
+		c.TaskData.TargetNamespace = string(c.TaskData.ForbiddenNamespace)
+	} else {
+		c.TaskData.TargetNamespace = options.ResolveNamespace(c.TaskData.TargetTemplateNamespace, options.TestNamespace)
+	}
+
+	c.TaskData.Template = testtemplate.NewCirrosServerTinyTemplate().Build()
```
atiratree, line 55: This method seems too hard coded and hard to configure. Can we just pass these values (such as name, template) to the taskData to also represent what the test case tests?

I know it is hard to understand to begin with, but that is due to a need of setting the right namespaces (since they are generated).

modules/tests/test/copy_template_test.go
```diff
@@ -141,28 +99,49 @@ var _ = Describe("Copy template task", func() {
...
+			resultTemplateName := results["name"]
+			resultTemplateNamespace := results["namespace"]
+
+			newTemplate, err := f.TemplateClient.Templates(resultTemplateNamespace).Get(context.TODO(), resultTemplateName, v1.GetOptions{})
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(newTemplate).ToNot(Equal(nil), "new template should exists")
+
+			f.ManageTemplates(newTemplate)
+		},
+			table.Entry("no target template name specified", &testconfigs.CopyTemplateTestConfig{
```
atiratree, line 137: do we need an additional table? Can we customize the one above?

modules/tests/test/copy_template_test.go
```diff
@@ -43,95 +37,59 @@ var _ = Describe("Copy template task", func() {
...
 				})
 
-			newTemplate, err := f.TemplateClient.Templates(config.TaskData.TargetTemplateNamespace).Get(context.TODO(), newTemplateName, v1.GetOptions{})
+			results := r.GetResults()
+			resultTemplateName := results["name"]
+			resultTemplateNamespace := results["namespace"]
+
+			newTemplate, err := f.TemplateClient.Templates(resultTemplateNamespace).Get(context.TODO(), resultTemplateName, v1.GetOptions{})
 			Expect(err).ShouldNot(HaveOccurred())
 			Expect(newTemplate).ToNot(Equal(nil), "new template should exists")
```
atiratree, line 94: `BeNil` could be used instead

modules/tests/test/copy_template_test.go
```diff
@@ -141,28 +99,49 @@ var _ = Describe("Copy template task", func() {
...
-					Template:                cirrosTemplate,
-					SourceTemplateName:      cirrosTemplate.Name,
-					SourceTemplateNamespace: cirrosTemplate.Namespace,
-					TargetTemplateName:      newTemplateName,
-					TargetTemplateNamespace: f.TestNamespace,
-				},
-			}, cirrosTemplate.Namespace,
-			),
+				TaskData: testconfigs.CopyTemplateTaskData{},
+			}),
```
atiratree, line 103: it would be nice to have also tests for the namespace defaulting (when task param is `""` )

atiratree (commented):
implementation looks good

modules/tests/test/copy_template_test.go
```diff
@@ -61,7 +58,8 @@ var _ = Describe("Copy template task", func() {
 					LimitTestScope: NamespaceTestScope,
 				},
 				TaskData: testconfigs.CopyTemplateTaskData{
-					ForbiddenNamespace: openshiftNamespace,
+					ForbiddenNamespace: SystemTargetNS,
```
atiratree, line 61: why not just pass it to the TargetNamespace? In that case we would not need the ForbiddenNamespace variable and it would simplify the code

modules/tests/test/copy_template_test.go
```diff
@@ -61,7 +58,8 @@ var _ = Describe("Copy template task", func() {
 					LimitTestScope: NamespaceTestScope,
 				},
 				TaskData: testconfigs.CopyTemplateTaskData{
```
atiratree, line 60: we could also add a similar test but the source namespace would be unreachable

modules/tests/test/copy_template_test.go
```diff
@@ -79,67 +77,52 @@ var _ = Describe("Copy template task", func() {
...
 					DoNotSetTargetName: true,
+					Template:           testtemplate.NewCirrosServerTinyTemplate().Build(),
+				},
+			}),
+			table.Entry("no target namespaces specified", &testconfigs.CopyTemplateTestConfig{
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
+					ServiceAccount: CopyTemplateServiceAccountName,
+				},
+				TaskData: testconfigs.CopyTemplateTaskData{
+					DoNotSetTargetNamespace: true,
```
atiratree, line 124: would be nice to have the same for source namespace

modules/tests/test/copy_template_test.go
```diff
@@ -79,67 +77,52 @@ var _ = Describe("Copy template task", func() {
...
+				},
 			}),
 			table.Entry("should create template in different namespace", &testconfigs.CopyTemplateTestConfig{
 				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
 					ServiceAccount: CopyTemplateServiceAccountName,
 					LimitTestScope: ClusterTestScope,
 				},
 				TaskData: testconfigs.CopyTemplateTaskData{
-					TargetTemplateNamespace: openshiftNamespace,
+					TargetTemplateNamespace: TestTargetNS,
```
atiratree, line 106: both source and target namespace default to the test one right? In that case DeployTargetNS should be here to create a template in a different namespace.

modules/tests/test/testconfigs/copy-template-config.go
```diff
@@ -46,14 +44,16 @@ func (c *CopyTemplateTestConfig) Init(options *testoptions.TestOptions) {
...
-	} else {
-		c.TaskData.TargetNamespace = options.ResolveNamespace(c.TaskData.TargetTemplateNamespace, options.TestNamespace)
+	if !c.TaskData.DoNotSetTargetNamespace {
+		if c.TaskData.ForbiddenNamespace != "" {
+			c.TaskData.TargetNamespace = string(c.TaskData.ForbiddenNamespace)
+		} else {
+			c.TaskData.TargetNamespace = options.ResolveNamespace(c.TaskData.TargetTemplateNamespace, options.TestNamespace)
+		}
+	}
+	if c.TaskData.Template != nil {
```
atiratree, line 54: +1 easier to read

modules/tests/test/testconfigs/copy-template-config.go
```diff
@@ -32,8 +32,6 @@ type CopyTemplateTestConfig struct {
 }
 
 func (c *CopyTemplateTestConfig) Init(options *testoptions.TestOptions) {
-	c.TaskData.ResultTemplateName = NewTemplateName
-
 	if c.TaskData.SourceTemplateName == "" && !c.TaskData.DoNotSetSourceName {
 		c.TaskData.SourceTemplateName = testtemplate.CirrosTemplateName
```
atiratree, line 36: this can be just set from the c.TaskData.Template if we have it no? And we can also pass the desired name through SourceTemplateName/TargetTemplateName in each test case . And there would not be a need for DoNotSetSourceName / DoNotSetTargetName.

Let me know if this makes sense?

atiratree (changes requested):

modules/tests/test/testconfigs/copy-template-config.go
```diff
@@ -32,28 +27,15 @@ type CopyTemplateTestConfig struct {
...
-	}
-
-	if c.TaskData.TargetTemplateName == "" && !c.TaskData.DoNotSetTargetName {
-		c.TaskData.TargetTemplateName = NewTemplateName
-	}
-
 	c.deploymentNamespace = options.DeployNamespace
 
-	c.TaskData.SourceNamespace = options.ResolveNamespace(c.TaskData.SourceTemplateNamespace, options.TestNamespace)
+	c.TaskData.SourceNamespace = options.ResolveNamespace(c.TaskData.SourceTemplateNamespace, options.DeployNamespace)
```
atiratree, line 32: originally having deploy/test namespace was done to test RBAC permissions. That is why we have tests: cluster-scope and namespace scope (where deploy == test)

so for the cluster scope we want to create taskruns in deploy ns and the entities they create in test ns. 
Because of that we should default to TestNamespace in our entities to test accross namespaces by default. 

Then we can have one or few tests that test the edge cases. Like creating entities in deploy namespace as mentioned here: https://github.com/kubevirt/kubevirt-tekton-tasks/pull/93#discussion_r708398066

atiratree: Although I am not sure if it still make sense to split it like that and have separate jobs for 2 sets of RBACs since it doesn't test the code but the deployment. Originally it wasn't clear if we would distribute two yamls (one for namespace scope and one for cluster scope) or just one, so these jobs were testing that. But since then we switched to distributing only the namespace scope and leaving the rest to the user: https://github.com/kubevirt/kubevirt-tekton-tasks/blob/main/docs/tasks-rbac-permissions.md

atiratree:
@ksimon1 looks good but it seems that we have a template collision: 

`templates.template.openshift.io "cirros-vm-template" already exists`

can you implement similar approach to this https://github.com/kubevirt/kubevirt-tekton-tasks/blob/main/modules/tests/test/testconfigs/dv-test-config.go#L39 ?

to always generate template names with random suffixes? This is also to prevent collisions with other tests.

atiratree:
in hindsight I should have probably made the testing infrastructure to generate new namespace for each test case so collisions could not happen. This would require quite a bit of effort now to change though..

atiratree (approved):

PR #96, opened 2021-08-05
add modify-vm-template task stub for ci

ksimon1:
add modify-vm-template task stub for ci

**Release note**:

```
NONE
```


configs/modify-vm-template-task.yaml
```diff
@@ -0,0 +1,3 @@
+task_name: modify-vm-template-task
+task_category: modify-vm-template-task
+main_image: quay.io/kubevirt/tekton-task-modify-vm-template-task:v0.0.1
```
atiratree, line 3: missing the okd part..

ksimon1: good catch, fixed

configs/modify-vm-template.yaml
```diff
@@ -0,0 +1,4 @@
+task_name: modify-vm-template
+task_category: modify-vm-template
+main_image: quay.io/kubevirt/tekton-modify-vm-template:v0.0.1
```
atiratree, line 3: I think this should have a `tekton-task` prefix as the other tasks. 

Are you using the onboard-new-task script?

PR #98, opened 2021-08-16
move log and output packages to shared module

ksimon1:
move log and output packages to shared module
these two packages will be used by copy-template task.

**Release note**:
```
NONE
```


atiratree:
@ksimon1 +1, we could later do this for other tasks as well

PR #111, opened 2022-02-10
regenerate local manifests with latest image tag

ksimon1:
regenerate local manifests with latest image tag
All containers will have same tag as release

update generating of local manifests so now all images are using the latest tag.
During release of manifests on github, images will have same tag as release name
(defined by variable IMG_TAG in file .github/workflows/release.yaml).
With this change containers will get regenerated during new release and
it will improve and simplify maintaining of images.

**Release note**:
```
NONE
```


scripts/generate-yaml-tasks.sh
```diff
@@ -34,6 +38,11 @@ function generateTaskResources() {
         DESTINATION_PARENT_DIR="${REPO_DIR}/tasks/${TASK_NAME}"
         DESTINATION="${DESTINATION_PARENT_DIR}/${RESOURCE_TYPE}"
         SOURCE="${REPO_DIR}/templates/${TASK_NAME}/dist/${RESOURCE_TYPE}"
+
```
atiratree, line 41: in case you would go this way, I think it is preferable to read an env variable in the ansible and set tag accordingly instead of doing post processing like this.

Also you can either remove the tag from configs or just create a new `main_image_tag` ansible variable there and then either use that or switch to the env variable accordingly.

scripts/generate-yaml-tasks.sh
```diff
@@ -26,6 +26,10 @@ else
   TASK_NAMES=("$@")
 fi
 
+if [ -z "${IMG_TAG}" ]; then
+  IMG_TAG="latest"
```
atiratree, line 30: where is this used?

makefile
```diff
@@ -1,3 +1,6 @@
+#current version of tekton tasks
+export VERSION ?=v0.0.9
```
atiratree, line 2: can we use something less generic? for example `RELEASE_VERSION` ?

scripts/deploy-tasks.sh
```diff
@@ -27,7 +27,7 @@ visit "${REPO_DIR}/tasks"
       continue
     fi
     CONFIG_FILE="${REPO_DIR}/configs/${TASK_NAME}.yaml"
-    MAIN_IMAGE="$(sed -n  's/^main_image *: *//p' "${CONFIG_FILE}")"
+    MAIN_IMAGE="$(sed -n  's/^main_image *: *//p' "${CONFIG_FILE}"):${VERSION}"
```
atiratree, line 30: can we guard against an empty value at the beginning of the script?

.github/workflows/release.yaml
```diff
@@ -45,3 +45,43 @@ jobs:
           asset_path: manifests/okd/kubevirt-tekton-tasks-okd.yaml
           asset_name: kubevirt-tekton-tasks-okd.yaml
           asset_content_type: text/plain
+      - name: Update tekton tasks manifests
+        run: |
+          # Define vars
+          export VERSION=${{ steps.get_release.outputs.tag_name }}
+          # If GITHUB_FORK_USER is changed, a new access token should be set as a repo secret (ACTIONS_TOKEN)
```
atiratree, line 52: is this necessary? Can't we just create a new branch in the main repo to use for the PR. Or just push to the main branch?

.github/workflows/release.yaml
```diff
@@ -45,3 +45,43 @@ jobs:
           asset_path: manifests/okd/kubevirt-tekton-tasks-okd.yaml
           asset_name: kubevirt-tekton-tasks-okd.yaml
           asset_content_type: text/plain
+      - name: Update tekton tasks manifests
```
atiratree, line 48: is this what we want? Updating the manifests after uploading them?

.github/workflows/release.yaml
atiratree: or alternatively we could just keep it simple and make a PR with regenerated manifests manually before the release?

I suppose we could do an opposite approach. Manually create a release after change in version is pushed to the main branch.

atiratree (commented):
- `MAIN_IMAGE` variable is used in `makefile-common.sh` that is used in each module `makefile`. Please either change it to a working state or remove this variable usage if you think it will not be needed anymore
- `onboard-new-task-with-ci-stub.sh` should be updated as well.

.github/workflows/release.yaml
ksimon1: Idea is, when new tag is released, this will create a PR with regenerated manifests with newest tag. I wanted to avoid manual interaction as much as possible.

ksimon1: Uploaded manifests (in github release) are already updated with newest version. This updates manifests in repo

scripts/deploy-tasks.sh
```diff
@@ -27,7 +31,7 @@ visit "${REPO_DIR}/tasks"
...
```
atiratree, line 34: RELEASE_VERSION

.github/workflows/release.yaml
atiratree: What I find a bit bothersome with this approach is that the repo is in inconsistent state until this PR merges, in contrast what the latest release points to. If you would build the images and create release on PR merge the only delay would be the automation compared to the manual merge.

But your approach is valid as well,  I will leave it to your judgement.

.github/workflows/release.yaml
```diff
@@ -45,3 +45,43 @@ jobs:
...
+        run: |
+          # Define vars
+          export VERSION=${{ steps.get_release.outputs.tag_name }}
```
atiratree, line 51: `RELEASE_VERSION`  and bellow as well V

ksimon1:
retest

.github/workflows/release.yaml
ksimon1: I understand your concerns, but in this position (when we have deploy manifests in repo) we have only two options:
1) manual releases as they are now
2) automation, which will have to do a update PR after release. 
From my point I am fine with option number 2, because it will have really small impact on user (user can use manifests which are in github release and are up-to-date). And if user would like to use newest code, user will have to build it from source code, which means automation will replace all tags to latest.

atiratree:
I think this would make it more straightforward: https://github.com/kubevirt/kubevirt-tekton-tasks/pull/111#discussion_r808327717
> Can't we just create a new branch in the main repo to use for the PR. Or just push to the main branch?

but otherwise LGTM

PR #123, opened 2022-04-21
Add missing defaults to modify-vm-template manifest

0xFelix:

The manifest was missing defaults for several parameters, which made it
necessary to declare empty values for them when invoking the task.

**Release note**:

```release-note
Add missing defaults for modify-vm-template task parameters
```


ksimon1:
Can you please add a test, which will check if these default values will not override the values in template?

PR #127, opened 2022-04-26
add param allowReplace in copy template task

ksimon1:
add Param allowReplace in copy template task
When this param is set, task will not fail, when template already
exists, but it will update already existing template with newly copied template

**Release note**:
```
add param allowReplace in copy template task
```

### DIFF
--- a/manifests/okd/kubevirt-tekton-tasks-okd.yaml
+++ b/manifests/okd/kubevirt-tekton-tasks-okd.yaml
@@ -140,6 +140,7 @@ metadata:
     targetTemplateName.params.task.kubevirt.io/kind: Template
     targetTemplateName.params.task.kubevirt.io/apiVersion: template.openshift.io/v1
     targetTemplateNamespace.params.task.kubevirt.io/type: namespace
+    allowReplace.params.task.kubevirt.io/type: boolean
   labels:
     task.kubevirt.io/type: copy-template
     task.kubevirt.io/category: copy-template
@@ -161,6 +162,10 @@ spec:
       description: Namespace of an target OpenShift template to create in. (defaults to active namespace)
       type: string
       default: ""
+    - name: allowReplace
+      description: Allow replacing already existing template (same combination name/namespace). Allowed values true/false
+      type: string
+      default: "false"
   results:
     - name: name
       description: The name of a template that was created.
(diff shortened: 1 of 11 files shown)


modules/tests/test/copy_template_test.go
```diff
@@ -164,4 +165,84 @@ var _ = Describe("Copy template task", func() {
...
+			f.ManageTemplates(t)
+
+			runner.NewTaskRunRunner(f, config.GetTaskRun()).
+				CreateTaskRun().
+				ExpectFailure().
+				ExpectLogs(config.GetAllExpectedLogs()...).
+				ExpectResults(nil)
+		})
+
+		FIt("taskrun succeeds and template is updated", func() {
```
0xFelix, line 203: FIt

tasks/copy-template/manifests/copy-template.yaml
```diff
@@ -31,6 +32,10 @@ spec:
       description: Namespace of an target OpenShift template to create in. (defaults to active namespace)
       type: string
       default: ""
+    - name: allowReplace
+      description: Allow replacing already existing template (same combination name/namespace). Allowed values true/false
+      type: string
+      default: ""
```
0xFelix, line 38: Same as above.

templates/copy-template/manifests/copy-template.yaml
```diff
@@ -31,6 +32,10 @@ spec:
       description: Namespace of an target OpenShift template to create in. (defaults to active namespace)
       type: string
       default: ""
+    - name: allowReplace
+      description: Allow replacing already existing template (same combination name/namespace). Allowed values true/false
+      type: string
+      default: ""
```
0xFelix, line 38: Same as above.

manifests/okd/kubevirt-tekton-tasks-okd.yaml
```diff
@@ -161,6 +162,10 @@ spec:
       description: Namespace of an target OpenShift template to create in. (defaults to active namespace)
       type: string
       default: ""
+    - name: allowReplace
+      description: Allow replacing already existing template (same combination name/namespace). Allowed values true/false
+      type: string
+      default: ""
```
0xFelix, line 168: Maybe default to "false"? But probably makes no difference.

PR #148, opened 2022-05-24
Update examples documentation

0xFelix:

Update docs for windows10-installer and add docs for
windows10-customize.

**Release note**:

```release-note
None
```


examples/pipelines/windows10-installer/README.md
```diff
@@ -0,0 +1,122 @@
+# Windows 10 Installer Pipeline
+
+Downloads a Windows Source ISO into a DV and installs Windows into a new base DV.
+It will then spin up an installation VM and use Windows Answer Files to automatically install the VM.
+Then the pipeline will wait for the installation to complete and will delete the installation VM while keeping the artifact DV with the installed operating system.
+The pipeline can be customized to support different installation requirements.
+
```
dominikholler, line 7: This pipeline installs Windows into a new DataVolume. This DataVolume is suitable to be used as a default boot source or golden image for Windows VMs.

The pipeline implements this by spinning up a new VM which boots from the Windows disc image (ISO file). The installation of windows is automatically executed and controlled by a Windows answer file. Then the pipeline will wait for the installation to complete and will delete the installation VM while keeping the resulting DataVolume with the installed operating system. The pipeline can be customized to support different installation requirements.

examples/pipelines/windows10-installer/README.md
```diff
@@ -0,0 +1,122 @@
...
+There is a specific version of this pipeline for OKD. This version is using templates, which are not available on Kubernetes.
+
+## Prerequisites
+
+- KubeVirt `v0.53.1`
```
dominikholler, line 12: exactly this version or at least this version?

examples/pipelines/windows10-installer/README.md
```diff
@@ -0,0 +1,122 @@
...
+4. Initialize a WIN_URL variable that will be used to create a DataVolume which will download this ISO into a PVC.
+
+```bash
+# Real URL can look differently
+WIN_URL="https://software.download.prss.microsoft.com/db/Win10_21H2_English_x64.iso..."
+```
+
+### Prepare autounattend.xml ConfigMap
+
+1. Supply, generate or use the default autounattend.xml.
```
dominikholler, line 36: Would be great if this information could be in https://kubevirt.io/user-guide/virtual_machines/startup_scripts/#sysprep, which could be referenced here.

dominikholler (changes requested):

examples/pipelines/windows10-installer/README.md
0xFelix: It only has to be a recent version of KubeVirt (can also be older). I chose this version as it was the latest version at the time of writing the docs.

examples/pipelines/windows10-installer/README.md
0xFelix: I changed the docs and created https://github.com/kubevirt/user-guide/pull/542

PR #152, opened 2022-06-09
Onboard create-data-object task

0xFelix:

This task will replace create-datavolume-from-manifest.
This time it is written in golang with its own task image.

**Release note**:

```release-note
Onboard create-data-object task
```


modules/create-datavolume-from-manifest/build/create-datavolume-from-manifest/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal AS builder
+RUN microdnf install -y golang-1.16.* && microdnf clean all
```
ksimon1, line 2: can you please bump golang to 1.18?

0xFelix: This is only for onboarding purposes. The golang version will be bumped in the PR with the actual changes.

PR #161, opened 2022-06-22
Add create-data-object task

### LINKED ISSUE #143: create-datavolume-from-manifest should be able to overwrite existing DV
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind enhancement

**What happened**:

If a pipelines tries to update an existing PV, the pipeline fails

**What you expected to happen**:

The pipeline can explicitly allow to overwrite or replace the DV

**How to reproduce it (as minimally and precisely as possible)**:

run https://github.com/kubevirt/tekton-tasks-operator/blob/main/data/tekton-pipelines/okd/windows10-installer.yaml twice.

0xFelix:

This task replaces create-datavolume-from-manifest task. It is written
in golang instead of bash and additionally supports the creation of
DataSources.

**Which issue(s) this PR fixes**:
Fixes #143 

**Release note**:

```release-note
Add create-data-object task
Remove create-datavolume-from-manifest task (replaced by create-data-object)
```


modules/create-data-object/build/create-data-object/Dockerfile
```diff
@@ -1,4 +1,19 @@
...
+WORKDIR /src/${TASK_NAME}
+RUN curl -L https://go.dev/dl/go1.18.3.linux-amd64.tar.gz | tar -C /usr/local -xzf -
+ENV PATH=$PATH:/usr/local/go/bin
+COPY . .
+RUN	CGO_ENABLED=0 GOOS=linux go build -o /${TASK_NAME} cmd/${TASK_NAME}/main.go
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+ENV TASK_NAME=create-data-object
+ENV ENTRY_CMD=/usr/local/bin/${TASK_NAME}
+
```
ksimon1, line 14: can you please add here command `microdnf update -y && microdnf clean all`

modules/create-data-object/pkg/constants/constants.go
```diff
@@ -0,0 +1,31 @@
+package constants
+
+import "time"
+
+// Exit codes
+const (
+	InvalidCLIInputExitCode    = -1
+	DataObjectCreatorErrorCode = 1
```
ksimon1, line 8: 2 got missing? :D

modules/create-data-object/build/create-data-object/Dockerfile
0xFelix: Is that really needed in the final image? I could not find it in other task Dockerfiles.

modules/create-data-object/pkg/constants/constants.go
0xFelix: The structure was borrowed from another task :D I changed it

modules/create-data-object/build/create-data-object/Dockerfile
ksimon1: This will update image, so we can prevent some vulnerabilities. I will add it to other images as well

0xFelix: Shouldn't the `latest` image already be up to date?

PR #162, opened 2022-06-23
Bump create-data-object golang version to 1.17 temporarily

0xFelix:

Bump create-data-object golang version to 1.17 temporarily

It is needed so the task can be onboarded to the CI.

**Release note**:

```release-note
None
```


ksimon1:
no need to wait for ci

PR #166, opened 2022-06-28
regenerate deploy manifests

ksimon1:
regenerate deploy manifests

**Release note**:
```
NONE
```


ksimon1:
this PR is no longer needed

PR #177, opened 2022-07-26
add support for pipelineRuns in testing framework

ksimon1:
This commit adds support for pipelineRuns and testing of pipelines into
testing framework.

**Release note**:
```
NONE
```

### DIFF
--- a/modules/tests/test/framework/framework.go
+++ b/modules/tests/test/framework/framework.go
@@ -21,12 +21,14 @@ var TestOptionsInstance = &testoptions.TestOptions{}
 var ClientsInstance = &clients.Clients{}
 
 type ManagedResources struct {
-	taskRuns    []*pipev1beta1.TaskRun
-	dataVolumes []*cdiv1beta1.DataVolume
-	dataSources []*cdiv1beta1.DataSource
-	vms         []*kubevirtv1.VirtualMachine
-	templates   []*templatev1.Template
-	secrets     []*corev1.Secret
+	taskRuns     []*pipev1beta1.TaskRun
+	pipelineRuns []*pipev1beta1.PipelineRun
+	pipelines    []*pipev1beta1.Pipeline
+	dataVolumes  []*cdiv1beta1.DataVolume
+	dataSources  []*cdiv1beta1.DataSource
+	vms          []*kubevirtv1.VirtualMachine
+	templates    []*templatev1.Template
+	secrets      []*corev1.Secret
 }
 type Framework struct {
 	*testoptions.TestOptions
@@ -94,6 +96,7 @@ func (f *Framework) TestSetup(config TestConfig) {
 func (f *Framework) AfterEach() {
 	failed := CurrentSpecReport().Failed()
 	taskRuns := f.managedResources.taskRuns
+	pipelineRuns := f.managedResources.pipelineRuns
 
 	if failed {
 		defer func() {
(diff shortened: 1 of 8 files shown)


modules/tests/test/framework/framework.go
```diff
@@ -118,7 +127,12 @@ func (f *Framework) AfterEach() {
 			defer f.TknClient.TaskRuns(taskRun.Namespace).Delete(context.TODO(), taskRun.Name, metav1.DeleteOptions{})
 		}
 	}
-
+	for _, pipelineRun := range pipelineRuns {
```
0xFelix, line 130: What about cleanup in failed state like above?

modules/tests/test/tekton/pipelinerun.go
```diff
@@ -0,0 +1,108 @@
...
+					defer wg.Done()
+					req := clients.CoreV1Client.Pods(pipelineNamespace).GetLogs(name, &v1.PodLogOptions{
+						Follow: true,
+					})
+					podLogs, err := req.Stream(context.TODO())
+					if err == nil {
+						defer podLogs.Close()
+						defer GinkgoRecover()
+
+						result, err := ioutil.ReadAll(podLogs)
```
0xFelix, line 75: Can be `io.ReadAll`

modules/tests/test/tekton/pipelinerun.go
```diff
@@ -0,0 +1,108 @@
...
+						pipelinePodsLogs.setLog(name, string(result))
+
+						gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+					}
+				}(pipelineRun.Namespace, podName)
```
0xFelix, line 80: Does this have a benefit over using the parameters directly in the closure?

modules/tests/test/tekton/pipelinerun.go
```diff
@@ -0,0 +1,108 @@
...
+					})
+					podLogs, err := req.Stream(context.TODO())
+					if err == nil {
+						defer podLogs.Close()
+						defer GinkgoRecover()
+
+						result, err := ioutil.ReadAll(podLogs)
+						pipelinePodsLogs.setLog(name, string(result))
+
+						gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
```
0xFelix, line 78: What if `req.Stream` returns an error? Shouldn't that be checked too?

modules/tests/test/testconfigs/pipeline-test-config.go
```diff
@@ -0,0 +1,46 @@
+package testconfigs
+
+import (
+	. "github.com/kubevirt/kubevirt-tekton-tasks/modules/tests/test/constants"
+	"github.com/kubevirt/kubevirt-tekton-tasks/modules/tests/test/framework/testoptions"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type PipelineRunData struct {
```
0xFelix, line 10: Do we want to support Workspaces in the future?

modules/tests/test/testconfigs/pipeline-test-config.go
```diff
@@ -0,0 +1,46 @@
...
+	Pipeline *v1beta1.Pipeline
+
+	deploymentNamespace string
+	PipelineRunData     PipelineRunData
+	PipelineRun         *v1beta1.PipelineRun
+}
+
+func (c *PipelineTestConfig) Init(options *testoptions.TestOptions) {
+	c.deploymentNamespace = options.DeployNamespace
+
```
0xFelix, line 28: Whitespace

modules/tests/test/tekton/pipelinerun.go
ksimon1: In case of error, the function will be triggered again. In this case I don't thing it has some value to log error.

modules/tests/test/testconfigs/pipeline-test-config.go
ksimon1: We can add support some time later as now there is no need for it

modules/tests/test/tekton/pipelinerun.go
0xFelix: I'm not sure about silently failing. It also ends the polling when getting the `PipelineRuns` fails. Also do we need to call `podLogs.Close` even if an error occured?

ksimon1: Without pipelineRun the script can't do anything, that is why its failing. 

I don't think, that we need to fail. The pod can be e.g. in starting phase (and thus logs unavailable). But in next iteration the connection can be successfull. I can add some loging, but failing the whole script is not neccessary.

We don't have to call `podLogs.Close` when error occured, because the `req.Stream` fn returns always `nil, err`, so podLogs will always be nil, when `err != nil`

0xFelix: Okay, then we can keep it like this. Maybe just add a comment so we know that it's okay to fail, like you wrote?

ksimon1: I added the comment

PR #202, opened 2022-11-07
allow setting of libguestfs paths from env variables

ksimon1:
allow setting of libguestfs paths from env variables
this will help to set different paths in ds

**Release note**:
```
NONE
```


modules/disk-virt-customize/pkg/execute/executor.go
```diff
@@ -24,17 +25,18 @@ func NewExecutor(clioptions *parse.CLIOptions, diskImagePath string) *Executor {
...
 
 func (e *Executor) PrepareGuestFSAppliance() error {
-	applianceArchivePath := constants.GuestFSApplianceArchivePath
+	applianceArchivePath := env.EnvOrDefault(constants.GuestFSApplianceArchivePathEnv, constants.GuestFSApplianceArchivePath)
 
 	if _, err := os.Stat(applianceArchivePath); os.IsNotExist(err) {
 		return zerrors.NewMissingRequiredError("guestfs appliance is missing at %v", applianceArchivePath)
 	}
 
+	targetFolder := env.EnvOrDefault(constants.TargetApplianceFolderEnv, "/home/disk-virt-customize")
```
0xFelix, line 34: Make the default a const too?

modules/disk-virt-sysprep/pkg/execute/executor.go
```diff
@@ -24,17 +25,18 @@ func NewExecutor(clioptions *parse.CLIOptions, diskImagePath string) *Executor {
...
 
 func (e *Executor) PrepareGuestFSAppliance() error {
-	applianceArchivePath := constants.GuestFSApplianceArchivePath
+	applianceArchivePath := env.EnvOrDefault(constants.GuestFSApplianceArchivePathEnv, constants.GuestFSApplianceArchivePath)
 
 	if _, err := os.Stat(applianceArchivePath); os.IsNotExist(err) {
 		return zerrors.NewMissingRequiredError("guestfs appliance is missing at %v", applianceArchivePath)
 	}
 
+	targetFolder := env.EnvOrDefault(constants.TargetApplianceFolderEnv, "/home/disk-virt-sysprep")
```
0xFelix, line 34: Make the default a const too?

modules/shared/pkg/env/env.go
```diff
@@ -31,3 +32,11 @@ func GetActiveNamespace() (string, error) {
 func GetTektonResultsDir() string {
 	return tektonResultsDirPath
 }
+
+func EnvOrDefault(envName string, defVal string) string {
+	val := os.Getenv(envName)
```
0xFelix, line 37: Use os.LookupEnv instead to make sure the variable was not set?

0xFelix (commented):
One more comment

PR #211, opened 2023-01-16
fix:remove namespace when processing template

ksimon1:
fix:remove namespace when processing template
In d/s e2e when trying to process template, api replied with panic
(panic: the namespace of the provided object does not match the
namespace sent on the request), because namespace send in request
is different than namespace in original template. This can be solved
by removing namespace from processed template and specifying only
namespace in request

**Release note**:
```
NONE

```


0xFelix (approved):
/lgtm

PR #218, opened 2023-02-06
Build and push ARM images

rgolangh:
Build and publish multi-arch images
    
    Motivation
    ARM clusters, specifically in the world of CI/CD are becoming
    increasingly interesting and kubevirt has added arm images for of its
    components. This would make those tekton tasks consumable in thoe
    clusters.
    
    Modification
    Build and push ARM images using the qemu-user-static emulation.
    Podman build supports '--platform' switch that makes the build choose
    the qemu emulation and performs the commands inside it.
    
    Result
    An image for both arm(aarch64) and amd64(existing) and a manifest
    which contains references to those multiple architectures.

Fixes #

I'm not adding anytihing into the release notes because I'm not sure this is 
used in an upstream release.
I forked this repo and tested those changes on the fork and published the results to 
quay.io/rgolangh/tekton-task-create-vm:v0.0.1 so you can inspect the manifest and 
see its multi arch

**Release note**:

```release-note

```

[CNV-24917](https://issues.redhat.com//browse/CNV-24917)

### DIFF
--- a/modules/copy-template/build/copy-template/Dockerfile
+++ b/modules/copy-template/build/copy-template/Dockerfile
@@ -1,13 +1,8 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y tar gzip && microdnf clean all
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18.9 AS builder
 ENV TASK_NAME=copy-template \
     GOFLAGS="-mod=vendor" \
     GO111MODULE=on
 WORKDIR /src/${TASK_NAME}
-
-RUN curl -L https://go.dev/dl/go1.19.2.linux-amd64.tar.gz | tar -C /usr/local -xzf -
-
-ENV PATH=$PATH:/usr/local/go/bin
 
 COPY . .
 RUN	CGO_ENABLED=0 GOOS=linux go build -o /${TASK_NAME} cmd/${TASK_NAME}/main.go
--- a/modules/create-vm/build/create-vm/Dockerfile
+++ b/modules/create-vm/build/create-vm/Dockerfile
@@ -1,11 +1,8 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y tar gzip && microdnf clean all
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18.9 AS builder
 ENV TASK_NAME=create-vm \
     GOFLAGS="-mod=vendor" \
     GO111MODULE=on
 WORKDIR /src/${TASK_NAME}
-RUN curl -L https://go.dev/dl/go1.19.2.linux-amd64.tar.gz | tar -C /usr/local -xzf -
-ENV PATH=$PATH:/usr/local/go/bin
 COPY . .
 RUN	CGO_ENABLED=0 GOOS=linux go build -o /${TASK_NAME} cmd/${TASK_NAME}/main.go
 
--- a/modules/disk-virt-customize/build/disk-virt-customize/Dockerfile
+++ b/modules/disk-virt-customize/build/disk-virt-customize/Dockerfile
@@ -1,11 +1,9 @@
 FROM registry.access.redhat.com/ubi9/ubi-minimal AS taskBuilder
-RUN microdnf install -y tar gzip && microdnf clean all
+RUN microdnf install -y tar gzip golang && microdnf clean all
 ENV TASK_NAME=disk-virt-customize \
     GOFLAGS="-mod=vendor" \
     GO111MODULE=on
 WORKDIR /src/${TASK_NAME}
-RUN curl -L https://go.dev/dl/go1.19.2.linux-amd64.tar.gz | tar -C /usr/local -xzf -
-ENV PATH=$PATH:/usr/local/go/bin
 COPY . .
 RUN	CGO_ENABLED=0 GOOS=linux go build -o /${TASK_NAME} cmd/${TASK_NAME}/main.go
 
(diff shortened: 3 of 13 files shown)


modules/copy-template/build/copy-template/Dockerfile
```diff
@@ -1,14 +1,10 @@
 FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y tar gzip && microdnf clean all
+RUN microdnf install -y tar gzip golang && microdnf clean all
```
ksimon1, line 2: we download golang from official site, because many times the newest release is not available in ubi image.

rgolangh: then maybe better to use a golang builder image like "registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13 AS builder "?  (just an example, could use newer )

rgolangh: this should save the the special code for arch aware downloads. 
Those tasks are fairly small and short lived commands and they should be okay not to use the latest golang release. Unless you are waiting for a new feature?

rgolangh: also noticed I could remove the installation of tar and gzip

ksimon1: I like the approach with registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.13 base image.

rgolangh:
+1

modules/copy-template/build/copy-template/Dockerfile
rgolangh: registry.ci seems authenticated. There is an open, unauthenticated and maintained go builder based on ubi already
which is multi-arch and its a great drop-in replacement - registry.access.redhat.com/ubi9/go-toolset - I'll use that

ksimon1: sounds good

ksimon1:
@rgolangh 
the tests are failing with:
```
[1/2] STEP 6/6: RUN CGO_ENABLED=0 GOOS=linux go build -o /${TASK_NAME} cmd/${TASK_NAME}/main.go
go build command-line-arguments: copying /tmp/go-build2582675590/b001/exe/a.out: open /copy-template: permission denied
error: build error: error building at STEP "RUN CGO_ENABLED=0 GOOS=linux go build -o /${TASK_NAME} cmd/${TASK_NAME}/main.go": error while running runtime: exit status 1
```

modules/copy-template/build/copy-template/Dockerfile
```diff
@@ -1,14 +1,9 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y tar gzip && microdnf clean all
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18.9 AS builder
```
0xFelix, line 1: The `go.mod` file of `copy-template` task has already moved to `go 1.19`, so using this older version as the builder can fail tthe build when using 1.19 features. Can't the builder use 1.19 too?

modules/create-vm/build/create-vm/Dockerfile
```diff
@@ -1,11 +1,8 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y tar gzip && microdnf clean all
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18.9 AS builder
```
0xFelix, line 1: Same as above.

modules/disk-virt-customize/build/disk-virt-customize/Dockerfile
```diff
@@ -1,11 +1,9 @@
 FROM registry.access.redhat.com/ubi9/ubi-minimal AS taskBuilder
-RUN microdnf install -y tar gzip && microdnf clean all
+RUN microdnf install -y tar gzip golang && microdnf clean all
```
0xFelix, line 2: Similar to above, how do we make sure the golang version matches the enabled golang feature set in go.mod?

modules/disk-virt-sysprep/build/disk-virt-sysprep/Dockerfile
```diff
@@ -1,11 +1,9 @@
 FROM registry.access.redhat.com/ubi9/ubi-minimal AS taskBuilder
-RUN microdnf install -y tar gzip && microdnf clean all
+RUN microdnf install -y tar gzip golang && microdnf clean all
```
0xFelix, line 2: Same as above.

modules/execute-in-vm/build/execute-in-vm/Dockerfile
```diff
@@ -1,11 +1,8 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y  tar gzip && microdnf clean all
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18.9 AS builder
```
0xFelix, line 1: See above.

modules/generate-ssh-keys/build/generate-ssh-keys/Dockerfile
```diff
@@ -1,11 +1,8 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal AS builder
-RUN microdnf install -y tar gzip && microdnf clean all
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18.9 AS builder
```
0xFelix, line 1: See above.

0xFelix (commented):
I have concerns about mismatches between the enabled golang features and the used golang version for the build. IMO we need to keep this in sync.

PR #220, opened 2023-02-09
fix: delete pvc when datavolume is not found

ksimon1:
fix: delete pvc when datavolume is not found
Datavolumes are garbage collected, when they are successfully imported. This means that when allowReplace param is set and datavolume is GC and pvc exists, modify-data-object did not delete pvc. This PR changes the behaviour so MDO task tries to delete the existing pvc, when datavolume does not exists.

**Release note**:
```
fix: delete pvc when datavolume is not found
```

### DIFF
--- a/modules/modify-data-object/pkg/dataobject/provider.go
+++ b/modules/modify-data-object/pkg/dataobject/provider.go
@@ -36,6 +36,7 @@ type DataObjectProvider interface {
 	GetPVC(string, string) (*v1.PersistentVolumeClaim, error)
 	DeleteDS(string, string) error
 	DeleteDV(string, string) error
+	DeletePVC(string, string) error
 	CreateDo(*unstructured.Unstructured, bool) (*unstructured.Unstructured, error)
 }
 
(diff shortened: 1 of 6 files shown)


modules/modify-data-object/pkg/dataobject/provider.go
```diff
@@ -66,6 +67,34 @@ func (d *dataObjectProvider) DeleteDS(namespace string, name string) error {
...
+		}
+
+		if err != nil {
+			return err
+		}
+
+		return d.DeletePVC(namespace, name)
+	}
+
+	if _, err := helper.Delete(namespace, name); err != nil {
```
0xFelix, line 92: Do it like this to avoid the additional if-statement?

```
_, err := helper.Delete(namespace, name)
return err
```

modules/tests/test/modify_data_object_test.go
```diff
@@ -218,6 +220,67 @@ var _ = Describe("Modify data objects", func() {
 			Expect(dv2.Spec).To(Equal(dv.Spec))
 		})
 
+		It("Existing pvc is deleted and replaced", func() {
+			const (
+				initialURL  = "https://invalid.url.initial"
```
0xFelix, line 225: Where is this used?

ksimon1: fixed

0xFelix (approved):
/lgtm
/retest

PR #225, opened 2023-02-17
feat: add delete PVC option for modify-data-object task

ksimon1:
feat: add delete PVC option for modify-data-object task
This PR adds ability to delete PVC via modify-data-object task. This will help during efi pipeline, because when the imported PVC is mounted to modify-windows-iso pod, it can't be deleted, because the PVC contains finalizer to prevent dataloss when mounted to pod. We have to include the delete pvc step into efi pipeline, so the pvc can be moved to terminating state and deleted when modify-windows-iso pod is deleted

**Release note**:
```
feat: modify-data-object task can delete PVC

```

### DIFF
--- a/modules/modify-data-object/pkg/dataobject/provider.go
+++ b/modules/modify-data-object/pkg/dataobject/provider.go
@@ -154,13 +154,15 @@ func NewDataObjectCreator(cliOptions *parse.CLIOptions) (*DataObjectCreator, err
 }
 
 func (d *DataObjectCreator) DeleteDataObject() error {
-	if d.cliOptions.DeleteObjectKind == constants.DataVolumeKind {
+	switch d.cliOptions.DeleteObjectKind {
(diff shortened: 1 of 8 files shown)


modules/modify-data-object/pkg/dataobject/provider.go
```diff
@@ -161,6 +161,11 @@ func (d *DataObjectCreator) DeleteDataObject() error {
 	if d.cliOptions.DeleteObjectKind == constants.DataSourceKind {
 		return d.dataObjectProvider.DeleteDS(d.cliOptions.DataObjectNamespace, d.cliOptions.DeleteObjectName)
 	}
+
```
0xFelix, line 164: You could make a switch/case out of this func.

0xFelix (approved):
/lgtm

One nit, feel free to fix.

PR #235, opened 2023-03-23
Add deprecation annotations

opokornyy:

This PR adds deprecation annotation to the tasks that will be deprecated and removed in future versions.

**Release note**:
```
None
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ Visit [RBAC permissions for running the tasks](docs/tasks-rbac-permissions.md) i
 
 ## Usage
 
+#### copy-template, modify-vm-template and create-vm-from-template tasks are deprecated and will be removed in future versions. These tasks based on templates will be replaced with create-vm task with enhancements related to instance types.
+
 #### Create Virtual Machines
 
 - [create-vm-from-manifest](tasks/create-vm-from-manifest)
--- a/templates/modify-vm-template/manifests/modify-vm-template.yaml
+++ b/templates/modify-vm-template/manifests/modify-vm-template.yaml
@@ -3,6 +3,7 @@ apiVersion: tekton.dev/v1beta1
 kind: ClusterTask
 metadata:
   annotations:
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -38,6 +38,8 @@ Visit [RBAC permissions for running the tasks](docs/tasks-rbac-permissions.md) i
 
 ## Usage
 
+#### copy-template, modify-vm-template and create-vm-from-template tasks are  deprecated and will be removed in future versions.
```
ksimon1, line 41: please remove the extra space between `are  deprecated `

ksimon1 (changes requested):
please use the https://www.conventionalcommits.org/en/v1.0.0/ for commits

README.md
opokornyy: Done

README.md
```diff
@@ -38,6 +38,8 @@ Visit [RBAC permissions for running the tasks](docs/tasks-rbac-permissions.md) i
 
 ## Usage
 
+#### copy-template, modify-vm-template and create-vm-from-template tasks are deprecated and will be removed in future versions.
```
0xFelix, line 41: Can we name a replacement yet? Or what's the plan going forward?

ksimon1: The plan is to deprecate template based tasks and replace them with create-vm-from-manifest with few enhancements related to instance types

templates/modify-vm-template/manifests/modify-vm-template.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: tekton.dev/v1beta1
 kind: ClusterTask
 metadata:
   annotations:
+    task.kubevirt.io/deprecated: "true"
```
ksimon1, line 6: can you please change the annotation to:
```
tekton.dev/deprecated: "true"
```

README.md
0xFelix: Can we add this to the readme to give more context?

PR #244, opened 2023-04-25
Add virtctl to tekton task

opokornyy:

This PR adds functionality to create-vm-from-manifest task to allow user to create vm with virtctl command.

**Release note**:
```None

```


modules/create-vm/cmd/create-vm/main.go
```diff
@@ -51,8 +50,8 @@ func main() {
 	if err := vmCreator.OwnVolumes(vm); err != nil {
 		exit.ExitFromError(OwnVolumesErrorExitCode, err)
 	}
-	runStrategy := cliOptions.GetRunStrategy()
```
0xFelix, line 54: Why drop this check?

modules/create-vm/go.mod
```diff
@@ -7,18 +7,20 @@ require (
...
 	github.com/openshift/api v0.0.0
 	github.com/openshift/client-go v0.0.0
+	github.com/spf13/cobra v1.5.0
 	go.uber.org/zap v1.19.1
 	k8s.io/api v0.25.2
 	k8s.io/apimachinery v0.25.2
 	k8s.io/client-go v12.0.0+incompatible
-	kubevirt.io/api v0.59.0
-	kubevirt.io/client-go v0.59.0
+	kubevirt.io/api v0.60.0-alpha.0
```
0xFelix, line 19: Can we stay on stable?

modules/create-vm/pkg/utils/parse/clioptions.go
```diff
@@ -35,6 +36,7 @@ type CLIOptions struct {
 	RunStrategy               string            `arg:"--run-strategy,env:RUN_STRATEGY" help:"Set run strategy to vm"`
 	Output                    output.OutputType `arg:"-o" placeholder:"FORMAT" help:"Output format. One of: yaml|json"`
 	Debug                     bool              `arg:"--debug" help:"Sets DEBUG log level"`
+	Virtctl                   string            `arg:"--virtctl,env:VIRTCTL" placeholder:"VIRTCTL" help:"Specifies a virtctl create vm params to be used to create VirtualMachine."`
```
0xFelix, line 39: `s/params/command/`?

modules/create-vm/pkg/utils/parse/clioptions_test.go
```diff
@@ -23,8 +23,8 @@ var _ = Describe("CLIOptions", func() {
 		Expect(err).Should(HaveOccurred())
 		Expect(err.Error()).To(ContainSubstring(expectedErrMessage))
 	},
-		Entry("no mode", "one of vm-manifest, template-name should be specified", &parse.CLIOptions{}),
-		Entry("both modes", "only one of vm-manifest, template-name should be specified", &parse.CLIOptions{
+		Entry("no mode", "only one of vm-manifest, template-name or virtctl should be specified", &parse.CLIOptions{}),
+		Entry("both modes", "only one of vm-manifest, template-name or virtctl should be specified", &parse.CLIOptions{
```
0xFelix, line 27: `s/both/multiple/`?

modules/create-vm/pkg/vm/vm-provider.go
```diff
@@ -21,9 +23,9 @@ func NewVirtualMachineProvider(client kubevirtcliv1.KubevirtClient) VirtualMachi
 }
 
 func (v *virtualMachineProvider) Create(namespace string, vm *kubevirtv1.VirtualMachine) (*kubevirtv1.VirtualMachine, error) {
-	return v.client.VirtualMachine(namespace).Create(vm)
+	return v.client.VirtualMachine(namespace).Create(context.TODO(), vm)
```
0xFelix, line 26: Are we ever going to change the context? Might use `context.Background()`?

modules/create-vm/pkg/vm/vm-provider.go
```diff
@@ -21,9 +23,9 @@ func NewVirtualMachineProvider(client kubevirtcliv1.KubevirtClient) VirtualMachi
...
 }
 
 func (v *virtualMachineProvider) Start(namespace, name string) error {
-	return v.client.VirtualMachine(namespace).Start(name, &kubevirtv1.StartOptions{})
+	return v.client.VirtualMachine(namespace).Start(context.TODO(), name, &kubevirtv1.StartOptions{})
```
0xFelix, line 30: Same.

modules/create-vm/pkg/vmcreator/vm-creator.go
```diff
@@ -80,10 +85,46 @@ func (v *VMCreator) CreateVM() (*kubevirtv1.VirtualMachine, error) {
...
 
+func (v *VMCreator) createVMVirtctl() (*kubevirtv1.VirtualMachine, error) {
+	var vm kubevirtv1.VirtualMachine
+	output := &bytes.Buffer{}
+
+	cmd := createCommand(v.cliOptions.Virtctl, output)
+
+	err := cmd.Execute()
+	if err != nil {
+		return nil, zerrors.NewSoftError("could not run virtctl: %v", err.Error())
```
0xFelix, line 102: `Failed to execute command`?

0xFelix: Can the error be wrapped with `%w`?

modules/create-vm/pkg/vmcreator/vm-creator.go
```diff
@@ -80,10 +85,46 @@ func (v *VMCreator) CreateVM() (*kubevirtv1.VirtualMachine, error) {
...
+	}
+
+	if err := yaml.Unmarshal([]byte(output.Bytes()), &vm); err != nil {
+		return nil, zerrors.NewSoftError("could not read from virtctl output: %v", err.Error())
```
0xFelix, line 106: Same.

modules/create-vm/pkg/vmcreator/vm-creator.go
```diff
@@ -80,10 +85,46 @@ func (v *VMCreator) CreateVM() (*kubevirtv1.VirtualMachine, error) {
...
+	}
+
+	namespace := v.targetNamespace
+	if namespace == "" {
+		if namespace, err = env.GetActiveNamespace(); err != nil {
+			return nil, zerrors.NewMissingRequiredError("can't get active namespace: %v", err.Error())
```
0xFelix, line 112: Same.

modules/create-vm/pkg/vmcreator/vm-creator.go
```diff
@@ -80,10 +85,46 @@ func (v *VMCreator) CreateVM() (*kubevirtv1.VirtualMachine, error) {
...
+		return v.createVMVirtctl()
 	}
 	return nil, zerrors.NewMissingRequiredError("unknown creation mode: %v", v.cliOptions.GetCreationMode())
 }
 
+func (v *VMCreator) createVMVirtctl() (*kubevirtv1.VirtualMachine, error) {
+	var vm kubevirtv1.VirtualMachine
+	output := &bytes.Buffer{}
+
+	cmd := createCommand(v.cliOptions.Virtctl, output)
```
0xFelix, line 98: Create command like it is done here? https://github.com/kubevirt/kubevirt/blob/3c25c8606e8aac1fcb31a73a3f444ebfffb27d20/tests/clientcmd/virtctl.go#L53

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +249,197 @@ var _ = Describe("Create VM from manifest", func() {
...
+
+		It("should succeed  without specified namespace", func() {
+			vmName := "my-vm-1"
+			config := &testconfigs.CreateVMTestConfig{
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
+					ServiceAccount: CreateVMFromManifestServiceAccountName,
+					ExpectedLogs:   ExpectedSuccessfulVMCreation,
+				},
+				TaskData: testconfigs.CreateVMTaskData{
+					Virtctl: fmt.Sprintf("--name %s --run-strategy=Halted", vmName),
```
0xFelix, line 291: Use same way to specify flags? Either with `=` or space.

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +249,197 @@ var _ = Describe("Create VM from manifest", func() {
...
+		It("should start with startVM set to true", func() {
+			vmName := "my-vm-3"
+			instancetypeName := "instancetype-3"
+			config := &testconfigs.CreateVMTestConfig{
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
+					ServiceAccount: CreateVMFromManifestServiceAccountName,
+					ExpectedLogs:   ExpectedSuccessfulVMCreation,
+				},
+				TaskData: testconfigs.CreateVMTaskData{
+					Virtctl: fmt.Sprintf("--name %s --run-strategy=Halted --instancetype %s", vmName, instancetypeName),
```
0xFelix, line 356: Same as above.

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +249,197 @@ var _ = Describe("Create VM from manifest", func() {
...
+			expectedVMStub := config.TaskData.GetExpectedVMStubMeta()
+			f.ManageVMs(expectedVMStub)
+
+			runner.NewTaskRunRunner(f, config.GetTaskRun()).
+				CreateTaskRun().
+				ExpectFailure().
+				ExpectLogs(config.GetAllExpectedLogs()...)
+		})
+
+		DescribeTable("should fail creating VM", func(config *testconfigs.CreateVMTestConfig) {
```
0xFelix, line 404: Isn't this more of a general test case that affects every creation mode? I would have placed it on a higher level.

0xFelix (commented):
Good start! I added some comments and one further question: Can the task return the name of the VM? This leaves it open to create a VM with a random name.

modules/create-vm/go.mod
ksimon1: Ondrej select this tag, because there was something missing in 0.59. @opokornyy can you please confirm?

opokornyy: Yes, the stable tag does not contain all of the virtctl functionality that is needed.

modules/create-vm/pkg/vmcreator/vm-creator.go
```diff
@@ -80,10 +85,46 @@ func (v *VMCreator) CreateVM() (*kubevirtv1.VirtualMachine, error) {
 		return v.createVMFromTemplate()
 	case constants.VMManifestCreationMode:
 		return v.createVMFromManifest()
+	case constants.VirtctlCreatingMode:
+		return v.createVMVirtctl()
 	}
 	return nil, zerrors.NewMissingRequiredError("unknown creation mode: %v", v.cliOptions.GetCreationMode())
 }
 
+func (v *VMCreator) createVMVirtctl() (*kubevirtv1.VirtualMachine, error) {
```
ksimon1, line 94: please divide this function to two, e.g.:
runVirtctl()
convertVM()

modules/create-vm/go.mod
0xFelix: For the latest fixes you potentially want to pin the version to a specific commit. E.g. to include https://github.com/kubevirt/kubevirt/pull/9651

modules/create-vm/cmd/create-vm/main.go
opokornyy: We can specify the `RunStrategy` either through virtctl parameters or directly in a manifest. The `runStrategy`  checked in this condition is specified separately in the task spec. If we don't use the `RunStrategy` parameter of a task the string will be empty and it will depend only on `StartVM` flag. Also `StartVM` should not perform any action if `runStrategy` of a created VM is already set to Always. - https://kubevirt.io/user-guide/virtual_machines/run_strategies/#virtctl

modules/create-vm/pkg/utils/parse/clioptions.go
opokornyy: In this task param should be specified only the parameters, that will be used with the virtctl command like this for example `--name vm-test --run-strategy Always`. So maybe it is beter to use params instead of a command?

opokornyy:
> Good start! I added some comments and one further question: Can the task return the name of the VM? This leaves it open to create a VM with a random name.

The task is just creating a VM and it does not return its name. It depends on a user if he specifies a name of a VM, if he does not for virtctl command, then the created VM will have a random name.

modules/create-vm/pkg/utils/parse/clioptions.go
0xFelix: If you could improve the help text that's fine.

modules/create-vm/cmd/create-vm/main.go
0xFelix: So this always tries to start the VM and mightt return an error from the API?

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +249,197 @@ var _ = Describe("Create VM from manifest", func() {
...
+
+			Expect(err).ShouldNot(HaveOccurred())
+
+			Expect(vm.Name).To(Equal(vmName))
+			Expect(*vm.Spec.RunStrategy).To(Equal(kubevirtv1.RunStrategyAlways))
+
+			deleteInstancetype(f, instancetypeName)
+		})
+
+		DescribeTable("should fail creating VM", func(config *testconfigs.CreateVMTestConfig) {
```
0xFelix, line 347: Merge this into `taskrun fails and no VM is created`?

0xFelix:
@opokornyy Tasks can have return values, so why not return the name/namespace of the created VM? Other tasks (e.g. creating DataVolumes) do the same.

modules/create-vm/cmd/create-vm/main.go
opokornyy: It only tries to start the VM when `StartVM` param is specified. It should not return error, when the VM is already running nothing happens.

opokornyy:
> @opokornyy Tasks can have return values, so why not return the name/namespace of the created VM? Other tasks (e.g. creating DataVolumes) do the same.

Yeah, I will take a look on that. 

@ksimon1 What do you thing about adding name/namespace of the VM as a return value of this task?

opokornyy:
> @opokornyy Tasks can have return values, so why not return the name/namespace of the created VM? Other tasks (e.g. creating DataVolumes) do the same.

I have to correct myself here, the task already have a return values with name/namespace of the VM that was created. 

https://github.com/kubevirt/kubevirt-tekton-tasks/blob/9cfaf988cde7cfc2a549d94414ea8834020224ba/templates/create-vm-from-manifest/manifests/create-vm.yaml#L85

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +286,136 @@ var _ = Describe("Create VM from manifest", func() {
...
+
+		It("should succeed without specified namespace", func() {
+			vmName := "my-vm-1"
+			config := &testconfigs.CreateVMTestConfig{
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
+					ServiceAccount: CreateVMFromManifestServiceAccountName,
+					ExpectedLogs:   ExpectedSuccessfulVMCreation,
+				},
+				TaskData: testconfigs.CreateVMTaskData{
+					Virtctl: fmt.Sprintf("--name %s --run-strategy Halted", vmName),
```
0xFelix, line 328: Use random name returned by task instead?

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +286,136 @@ var _ = Describe("Create VM from manifest", func() {
...
+		It("should succeed with instancetype specified", func() {
+			vmName := "my-vm-2"
+			instancetypeName := "instancetype-2"
+			config := &testconfigs.CreateVMTestConfig{
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
+					ServiceAccount: CreateVMFromManifestServiceAccountName,
+					ExpectedLogs:   ExpectedSuccessfulVMCreation,
+				},
+				TaskData: testconfigs.CreateVMTaskData{
+					Virtctl: fmt.Sprintf("--name %s --instancetype %s", vmName, instancetypeName),
```
0xFelix, line 358: Same as above.

modules/tests/test/create_vm_from_manifest_test.go
```diff
@@ -242,6 +286,136 @@ var _ = Describe("Create VM from manifest", func() {
...
+		It("should start with startVM set to true", func() {
+			vmName := "my-vm-3"
+			instancetypeName := "instancetype-3"
+			config := &testconfigs.CreateVMTestConfig{
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
+					ServiceAccount: CreateVMFromManifestServiceAccountName,
+					ExpectedLogs:   ExpectedSuccessfulVMCreation,
+				},
+				TaskData: testconfigs.CreateVMTaskData{
+					Virtctl: fmt.Sprintf("--name %s --run-strategy Halted --instancetype %s", vmName, instancetypeName),
```
0xFelix, line 392: Same as above.

modules/create-vm/cmd/create-vm/main.go
```diff
@@ -51,10 +49,10 @@ func main() {
...
-	if cliOptions.GetStartVMFlag() && kubevirtv1.RunStrategyAlways != kubevirtv1.VirtualMachineRunStrategy(runStrategy) {
-		err := vmCreator.StartVM(vm.Namespace, vm.Name)
-		if err != nil {
+
+	if cliOptions.GetStartVMFlag() &&
+		(vm.Spec.RunStrategy == nil || *vm.Spec.RunStrategy != kubevirtv1.RunStrategyAlways) {
```
0xFelix, line 54: What happens if `spec.running` is already `true`, i.e. the VM is already running?

modules/create-vm/go.mod
0xFelix: You might want to update this to `v1.0.0-alpha.0` to have https://github.com/kubevirt/kubevirt/pull/9651 included.

modules/create-vm/cmd/create-vm/main.go
opokornyy: It caused an error and should be fixed now.

0xFelix (approved):
/approve

PR #245, opened 2023-05-11
feat: clean up create-vm task

ksimon1:
feat: clean up create-vm task

This PR should clean up create-vm task. It removes parameters which are
no longer needed, because they can be set directly in manifest or virtctl
parameters.
Additional clean up is done in the code, where I am removing code, which
is no longer in use.
**Release note**:
```
 clean up create-vm task
```


ksimon1:
@akrejcir, @jcanocan, @codingben, @opokornyy can you please review?

PR #253, opened 2023-06-19
[release-v0.13] chore: debug wait-for-vmi and execute-in-vm tests

ksimon1:
chore: debug wait-for-vmi and execute-in-vm tests

This PR updates behaviour of WaitForTaskRunState function, which will
not fail after timeout, but instead it saves error to struct which can be then
recognized and run some after code (in this case print out status of VM)
**Release note**:
```
NONE
```

### DIFF
--- a/automation/e2e-deploy-resources.sh
+++ b/automation/e2e-deploy-resources.sh
@@ -6,8 +6,7 @@ if kubectl get namespace tekton-pipelines > /dev/null 2>&1; then
   exit 0
 fi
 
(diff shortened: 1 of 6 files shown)


0xFelix (commented):
Could we also do this with https://github.com/kubevirt/must-gather?

0xFelix:
Btw. this is no backport but wants to merge into `release-0.13`?

ksimon1:
no, because the error are only occuring in 4.13 yet

PR #254, opened 2023-07-03
chore: debug wait-for-vmi, disk-virt and execute-in-vm tests

ksimon1:
This PR updates behaviour of WaitForTaskRunState function, which will not fail after timeout, but instead it saves error to struct which can be then recognized and run some after code (in this case print out status of VM)

**Release note**:
```
NONE
```

### DIFF
--- a/modules/tests/test/disk_virt_libguestfs_tasks_test.go
+++ b/modules/tests/test/disk_virt_libguestfs_tasks_test.go
@@ -42,10 +42,19 @@ var _ = Describe("Run disk virt-customize / virt-sysprep", func() {
 				Expect(err).ShouldNot(HaveOccurred())
 			}
 
-			runner.NewTaskRunRunner(f, config.GetTaskRun()).
+			r := runner.NewTaskRunRunner(f, config.GetTaskRun()).
 				CreateTaskRun().
-				ExpectFailure().
-				ExpectLogs(config.GetAllExpectedLogs()...).
+				ExpectFailure()
+			// debug lines, which should help to debug flaky tests
+			err := r.GetError()
+			if err != nil {
+				r.PrintTektonInfo()
+				if dv := config.TaskData.Datavolume; dv != nil {
+					r.PrintDatavolumeInfo(config.TaskData.Datavolume.Namespace, config.TaskData.Datavolume.Name)
+				}
+				Fail("error should not have occured")
+			}
+			r.ExpectLogs(config.GetAllExpectedLogs()...).
 				ExpectResults(nil)
 		},
 			Entry("no pvc", &testconfigs.DiskVirtLibguestfsTestConfig{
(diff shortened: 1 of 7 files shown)


modules/tests/test/disk_virt_libguestfs_tasks_test.go
```diff
@@ -42,10 +42,19 @@ var _ = Describe("Run disk virt-customize / virt-sysprep", func() {
...
-				ExpectLogs(config.GetAllExpectedLogs()...).
+				ExpectFailure()
+			// debug lines, which should help to debug flaky tests
+			err := r.GetError()
+			if err != nil {
+				r.PrintTektonInfo()
+				if dv := config.TaskData.Datavolume; dv != nil {
+					r.PrintDatavolumeInfo(config.TaskData.Datavolume.Namespace, config.TaskData.Datavolume.Name)
+				}
+				Expect(err).ShouldNot(HaveOccurred())
```
0xFelix, line 55: When `err != nil` this will always fail?

modules/tests/test/execute_and_cleanup_vm_test.go
```diff
@@ -93,14 +94,30 @@ var _ = Describe("Execute in VM / Cleanup VM", func() {
...
+				ExpectSuccessOrFailure(config.ExpectSuccess)
+
+			// debug lines, which should help to debug flaky tests
+			err := r.GetError()
+			if err != nil {
+				r.PrintTektonInfo()
+				if vm := config.TaskData.VM; vm != nil {
+					r.PrintVMInfo(vm.Namespace, vm.Name)
+				}
+				Expect(err).ShouldNot(HaveOccurred())
```
0xFelix, line 117: Same as above.

modules/tests/test/execute_and_cleanup_vm_test.go
```diff
@@ -447,10 +464,21 @@ var _ = Describe("Execute in VM / Cleanup VM", func() {
...
+			ExpectSuccessOrFailure(config.ExpectSuccess)
+
+		// debug lines, which should help to debug flaky tests
+		err := r.GetError()
+		if err != nil {
+			r.PrintTektonInfo()
+			if vm := config.TaskData.VM; vm != nil {
+				r.PrintVMInfo(vm.Namespace, vm.Name)
+			}
+			Expect(err).ShouldNot(HaveOccurred())
```
0xFelix, line 478: Same

modules/tests/test/wait_for_vmi_status_test.go
```diff
@@ -30,10 +30,21 @@ var _ = Describe("Wait for VMI Status", func() {
...
+			ExpectSuccessOrFailure(config.ExpectSuccess)
+
+		// debug lines, which should help to debug flaky tests
+		err := r.GetError()
+		if err != nil {
+			r.PrintTektonInfo()
+			if vm := config.TaskData.VM; vm != nil {
+				r.PrintVMInfo(vm.Namespace, vm.Name)
+			}
+			Expect(err).ShouldNot(HaveOccurred())
```
0xFelix, line 44: Same

modules/tests/test/disk_virt_libguestfs_tasks_test.go
ksimon1: yes, because if there is an error, it will anyway fail in later steps

0xFelix: Call `Fail()` instead to make it obvious failing is intended if this line is reached?

PR #258, opened 2023-08-10
fix: custom configuration of crio in testing cluster

ksimon1:
fix: custom configuration of crio in testing cluster

our e2e are not working due to permission issue in crio. cri-o/cri-o#7192 should revert the change. Until that revert PR is released and available in ci envs, we have to set special configuration on crio on all nodes

**Release note**:
```
NONE
```


ksimon1:
this PR is no longer needed

PR #261, opened 2023-09-15
fix: deployment script for minikube

ksimon1:
fix: deployment script for minikube

**Release note**:
```
NONE

```


ksimon1:
/hold 
I need to do more changes

PR #262, opened 2023-09-20
fix: change the condition to consider a DV import failed

jcanocan:
The test `TaskRun fails and DataVolume is modified but does not import successfully` checks if a DV has failed by asserting several fields on the DV. One of them is `conditions[cdiv1beta1.DataVolumeBound].Status == v1.ConditionTrue`. In the current version of CDI, if a DV fails to fetch the import link, which is the scenario in this test, the underlining PVC is not bound. Therefore, the test assertion fails even if the DV has failed to import.

It changes the logic of the function `HasDataVolumeFailedToImport` to check if a DV import has failed. Drops assert `conditions[cdiv1beta1.DataVolumeBound].Status == v1.ConditionTrue`.

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/modules/tests/test/dataobject/operations.go
+++ b/modules/tests/test/dataobject/operations.go
@@ -56,7 +56,6 @@ func HasDataVolumeFailedToImport(dataVolume *cdiv1beta1.DataVolume) bool {
 	conditions := getConditionMapDv(dataVolume)
 	return dataVolume.Status.Phase == cdiv1beta1.ImportInProgress &&
 		dataVolume.Status.RestartCount > constants.UnusualRestartCountThreshold &&
-		conditions[cdiv1beta1.DataVolumeBound].Status == v1.ConditionTrue &&
 		conditions[cdiv1beta1.DataVolumeRunning].Status == v1.ConditionFalse &&
 		conditions[cdiv1beta1.DataVolumeRunning].Reason == constants.ReasonError
 }


modules/tests/test/dataobject/operations.go
```diff
@@ -56,7 +56,7 @@ func HasDataVolumeFailedToImport(dataVolume *cdiv1beta1.DataVolume) bool {
 	conditions := getConditionMapDv(dataVolume)
 	return dataVolume.Status.Phase == cdiv1beta1.ImportInProgress &&
 		dataVolume.Status.RestartCount > constants.UnusualRestartCountThreshold &&
-		conditions[cdiv1beta1.DataVolumeBound].Status == v1.ConditionTrue &&
+		conditions[cdiv1beta1.DataVolumeBound].Status == v1.ConditionFalse &&
```
0xFelix, line 59: Can the check be dropped instead? Do the other conditions suffice to assert the import failed?

jcanocan: Yeah, probably the rest of the asserts are enough to check if the DV has failed. WDYT @ksimon1?

ksimon1: fine by me

jcanocan: Done!

0xFelix (approved):
/approve

Thanks!

PR #291, opened 2023-11-30
feat: configure dependabot

ksimon1:
this PR adds configuration for dependabot for multiple branches, adds assignees to PRs and allow bumping of all dependencies (direct, indirect)
/cc @0xFelix 
**Release note**:
```
NONE
```


ksimon1:
merging without tests, because this PR does not affect them

PR #297, opened 2023-12-12
fix: replace include with include_tasks

ksimon1:
fix: replace include with include_tasks

ansible.builtin.include has been deprecated and removed from ansible-core. Because of that, release scripts are not working - https://github.com/kubevirt/kubevirt-tekton-tasks/actions/runs/7183095492/job/19561070924

**Release note**:
```
NONE
```


0xFelix (approved):
/approve

Thanks!

PR #305, opened 2024-01-02
fix: race condition when allowReplace is true in modify-data-object task

ksimon1:
fix: race condition when allowReplace is true in modify-data-object task

modify-data-object is not waiting until dv/ds/pvc is deleted when allowReplace is set to true and it
tries to recreate the object. Sometimes the object still exists and task run ends with error that object still exists. This commit adds waiting functions which will make sure object no longer exists

**Release note**:
```
fix: race condition when allowReplace is true in modify-data-object task
```


jcanocan:
Nice job @ksimon1! Thanks!
/lgtm

PR #306, opened 2024-01-04
fix: generator of object names

ksimon1:
fix: generator of object names
Sometimes the generator returned the same hash which led to errors in tests. This commit changes the generator to use the remainder after division of random number to generate 5 char hash

**Release note**:
```
NONE
```


modules/sharedtest/testconstants/names.go
```diff
@@ -1,42 +1,23 @@
...
-
-	// xor highest 12 bits with lower ones and keep only lower bits
-	hash32 = ((hash32 ^ (hash32 >> bitCount)) << clearBitCount) >> clearBitCount
+const chars = "abcdefghijklmnopqrstuvwxyz1234567890"
 
-	// forcefully add left fifth digit if lower number
-	lowest5DigitNum := uint32(1) << (bitCount - 4)
-	if hash32 < lowest5DigitNum {
-		hash32 ^= lowest5DigitNum
+func fiveDigitTestHash() string {
```
akrejcir, line 10: To generate a random string, you can use the `rand.String()` function from `k8s.io/apimachinery/pkg/util/rand` package.

The `k8s.io/apimachinery` is already imported in this `sharedtest` package.

jcanocan:
This is great! There's still a chance to get collisions. But this should be enough for this purpose. If this is really a problem, consider to increase the number of bytes.
/lgtm

modules/sharedtest/testconstants/names.go
ksimon1: fixed

PR #316, opened 2024-01-18
feat: configure dependabot

ksimon1:
remove configuration for release branches, keep config only for main branch. Reduce number of PRs to max 3.
Enable grouping of dependencies to a single PR.
Add release-note-none label

**Release note**:
```
NONE
```


0xFelix (approved):
/lgtm

PR #319, opened 2024-01-30
feat: remove cluster wide permission objects

ksimon1:
feat: remove cluster wide permission objects

Kubevirt-tekton-tasks does not need to ship additional service accounts, role bindings and cluster roles, because if no service account is provided, tekton uses default SA pipeline, which is available in almost all namespaces and already has all necessary permission. This change will allow non privilidged user to run tasks without need admin permissions. The permission objects will remain only in README of each task so user can easily make it work if user will need multinamespace functionality.

**Release note**:
```
feat: remove cluster wide permission objects
```


manifests/kubernetes/kubevirt-tekton-tasks-kubernetes.yaml
```diff
@@ -795,60 +617,10 @@ spec:
...
-  name: modify-data-object-task
-  apiGroup: rbac.authorization.k8s.io
-subjects:
-  - kind: ServiceAccount
-    name: modify-data-object-task
 ---
 apiVersion: tekton.dev/v1
 kind: Task
 metadata:
-  annotations:
```
0xFelix, line 850: Btw. where was this account defined before? I do not see it in this file.

manifests/okd/kubevirt-tekton-tasks-okd.yaml
```diff
@@ -1237,47 +902,10 @@ spec:
...
-  apiGroup: rbac.authorization.k8s.io
-subjects:
-  - kind: ServiceAccount
-    name: modify-vm-template-task
 ---
 apiVersion: tekton.dev/v1
 kind: Task
 metadata:
-  annotations:
-    task.kubevirt.io/associatedServiceAccount: modify-windows-iso-file-task
```
0xFelix, line 1280: Same.

modules/tests/test/copy_template_test.go
```diff
@@ -94,66 +81,39 @@ var _ = Describe("Copy template task", func() {
...
-					ServiceAccount: CopyTemplateServiceAccountName,
-				},
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{},
 				TaskData: testconfigs.CopyTemplateTaskData{
 					SourceTemplateName: testtemplate.CirrosTemplateName,
 					TargetTemplateName: NewTemplateName,
 					Template:           testtemplate.NewCirrosServerTinyTemplate().Build(),
 				},
 			}),
-			Entry("should create template in different namespace", &testconfigs.CopyTemplateTestConfig{
```
0xFelix, line 106: Test no longer required?

modules/tests/test/copy_template_test.go
```diff
@@ -94,66 +81,39 @@ var _ = Describe("Copy template task", func() {
...
 				},
 			}),
 			Entry("no namespaces specified", &testconfigs.CopyTemplateTestConfig{
-				TaskRunTestConfig: testconfigs.TaskRunTestConfig{
-					ServiceAccount: CopyTemplateServiceAccountName,
-				},
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{},
 				TaskData: testconfigs.CopyTemplateTaskData{
-					SourceTemplateName:      testtemplate.CirrosTemplateName,
-					SourceTemplateNamespace: EmptyTargetNS,
```
0xFelix, line 153: Why are all namespaces not required anymore? No permissions?

scripts/ansible/generate-roles.yaml
```diff
@@ -22,7 +22,7 @@
     src: "{{ role_output_dir }}/tmp-role"
     dest: "{{ role_output_dir }}/{{ item.prefix if item.prefix is defined else '' }}{{ task_name }}-rbac.yaml"
     mode: "{{ default_file_mode }}"
-    delimiter: '\n'
```
0xFelix, line 25: Why change the delimiter?

0xFelix (commented):
IIUC this allows users to run the tasks without special permissions, but limits them to their namespace. In this case we should a note making that clear? Would it make sense to retain the cross namespace functionality, so cluster admins would still be able to grant the required permissions if they want to?

modules/tests/test/copy_template_test.go
ksimon1: yes

manifests/kubernetes/kubevirt-tekton-tasks-kubernetes.yaml
ksimon1: it is a bug. The SA is defined here, but in reality it never existed :D

ksimon1:
> IIUC this allows users to run the tasks without special permissions, but limits them to their namespace. In this case we should a note making that clear? Would it make sense to retain the cross namespace functionality, so cluster admins would still be able to grant the required permissions if they want to?

In case anybody would like to use multi namespace use case, there is a special section '### Usage in different namespaces' in task readme, where are all necessary objects and permissions

codingben:
@ksimon1 Will CI be able to test the changes properly? I mean to make it's deployable & runnable without being admin user in your local cluster. Also, will it require some changes in `ssp-operator`, or do you plan to remove it from this operator later?

templates/copy-template/generate-task.yaml
```diff
@@ -12,21 +12,27 @@
...
+    - name: Prepare readme dist directory
+      file:
+        path: "{{ item }}"
+        state: directory
+      with_items:
+        - "{{ readme_output_dir_tmp }}"
+    - name: Generate roles
+      include_tasks: "{{ repo_dir }}/scripts/ansible/generate-roles.yaml"
+      with_items:
+        - { role_type: ClusterRole, prefix: zz- }
```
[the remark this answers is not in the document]
ksimon1: because then you can link it from each namespace and you don't have to create role in each namespace. These cluster roles will not be part of manifests anymore. They will be only part of readme to help users if they will require multinamespace usecase. User can then select if they want to use role or clusterrole.

.github/workflows/test-yaml-consistency.yaml
0xFelix, line 1: I'm not the biggest fan of maintaining multiple CI systems. In other repos we have a generic Preow job, which run things like consistency checks and other linters. If we decide to keep this, please put it into a separate commit at least.

automation/e2e-deploy-resources.sh
0xFelix, line 1: Put these changes into a separate commit too please.

scripts/ansible/generate-roles.yaml
0xFelix: Can you create a separate commit explaining the need for this?

scripts/ansible/init-task-generation.yaml
```diff
@@ -5,5 +5,5 @@
 - name: Prepare dist directory
   file:
     # includes also manifests_output_dir
-    path: "{{ manifests_output_dir }}/tmp"
```
0xFelix, line 8: Same, why is this changed?

scripts/cluster-sync.sh
0xFelix, line 1: This looks like a refactoring, which belongs into a separate PR?

scripts/generate-yaml-tasks.sh
0xFelix, line 1: What's the motivation for these changes?

0xFelix (commented):
This PR really changes a lot and the lack of multiple commits makes it really hard to review. Can you please create separate commits, explaining what is done and why it is needed? I see also a lot of refactorings, which strictly speaking do not belong into this PR. Can you please create other PRs for the refactorings and cleanups you found?

.github/workflows/test-yaml-consistency.yaml
ksimon1: we are not adding any new CI. We are already using GH actions for mltiple other tests in this repo. This change is simplyfing overcomplex script. Problem is, ocp is doing changes to cloned repository, which causes false positives for git status --porcelain. On the another hand on GH action it works without any problems. I agree with moving this to another commit. On the contrary introducing kubevirt prow jobs would be introducing new ci system to this repo :)

0xFelix: Kubevirt Prow is part of all repos in the kubevirt org, you would simply need to add a job definition. Nevertheless, I still think this should be a separate PR.

modules/tests/test/copy_template_test.go
```diff
@@ -241,27 +196,24 @@ var _ = Describe("Copy template task", func() {
...
-				},
+				TaskRunTestConfig: testconfigs.TaskRunTestConfig{},
 				TaskData: testconfigs.CopyTemplateTaskData{
-					SourceTemplateName:      testtemplate.CirrosTemplateName,
-					TargetTemplateName:      NewTemplateName,
-					TargetTemplateNamespace: TestTargetNS,
-					AllowReplace:            "true",
-					Template:                testtemplate.NewCirrosServerTinyTemplate().Build(),
+					SourceTemplateName: testtemplate.CirrosTemplateName,
+					TargetTemplateName: E2ETestsRandomName(NewTemplateName),
```
0xFelix, line 202: This adds `E2ETestsRandomName`, is that required anywhere else?

modules/tests/test/framework/testoptions/test-options.go
```diff
@@ -79,47 +69,13 @@ func InitTestOptions(testOptions *TestOptions) error {
...
-
-		ns := f.targetNamespaces[namespace]
-
-		if ns != "" {
-			return ns
-		}
-	}
-
-	return f.TestNamespace
+func (f *TestOptions) GetDefaultNamespace() string {
```
0xFelix, line 79:
```suggestion
func (f *TestOptions) GetDeployNamespace() string {
```

It's the namespace where everything was deployed, not necessarily the default one?

.github/workflows/test-yaml-consistency.yaml
ksimon1: https://github.com/kubevirt/kubevirt-tekton-tasks/pull/321

modules/tests/test/framework/testoptions/test-options.go
ksimon1: fixed

modules/tests/test/copy_template_test.go
ksimon1: I updated the code and added the suffix to each target template in copy-template config to prevent flakiness

0xFelix (commented):
/approve

PR #322, opened 2024-02-12
feat: update to Go 1.21

ksimon1:
Update all necessary files to use golang 1.21

**Release note**:
```
bump golang version to 1.21
```


.github/workflows/release.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: 1.19
+          go-version: 1.21
```
0xFelix, line 18: Make this read the go version from `go.mod` file instead?

ksimon1: Correct, I will fix it

0xFelix (approved):
/lgtm

PR #327, opened 2024-02-14
feat: unify of okd and k8s tasks

ksimon1:
feat: unify of okd and k8s tasks

this commit removes okd / kubernetes tasks manifest and keeps only a single manifest file with all tasks. This change will simplify releases and we will not have to generate two files.

**Release note**:
```
Unify okd and k8s tasks manifests

```


.github/workflows/release.yaml
```diff
@@ -29,26 +29,15 @@ jobs:
           export RELEASE_VERSION="${{ steps.get_release.outputs.tag_name }}"
           make release
 
-      - name: Upload kubernetes manifest
-        id: kubernetes-manifest 
+      - name: Upload manifest
+        id: manifest
```
jcanocan, line 33:
```suggestion
        id: manifest
```
nit: whitespace

jcanocan (commented):
Just a minimal nit, apart from that, looks good.

.github/workflows/release.yaml
ksimon1: fixed

0xFelix (commented):
Looks like `README.md` needs updates as well.

ksimon1:
> Looks like `README.md` needs updates as well.

fixed

0xFelix (approved):
/lgtm

Thanks!

PR #329, opened 2024-02-15
fix: remove test private ssh keys from manifests

ksimon1:
fix: remove test private ssh keys from manifests

remove all test private ssh keys from manifests, because they are triggering false positive warnings

**Release note**:
```
NONE
```


jcanocan:
Thanks!
/lgtm

PR #331, opened 2024-02-16
fix: tekton version in templates and tasks

ksimon1:
fix: tekton version in templates and tasks

fix tekton version to v1 in templates and tasks

**Release note**:
```
Use tekton v1 in taskruns
```


0xFelix (approved):
/approve
/lgtm

PR #332, opened 2024-02-20
[WIP] feat: Add windows pipelines

ksimon1:
**What this PR does / why we need it**:
feat: Add windows pipelines
This PR moves windows example pipelines from SSP operator https://github.com/kubevirt/ssp-operator/tree/master/data/tekton-pipelines. It also add automation for generating pipelines and e2e tests 

This PR is WIP

**Release note**:
```
Move example pipelines from ssp operator to kubevirt-tekton-tasks
```

PR #337, opened 2024-03-08
chore: Generate manifests from #332

ksimon1:
This PR regenerates manifests from #332

**Release note**:
```
NONE

```


0xFelix (commented):
Are these generated files required to be part of the repo? Can you explain?

ksimon1:
Yes, because artifact hub requires to have manifests as a part of repo.

0xFelix (approved):
/lgtm

LGTM for now, butt in the long run we should find a better way than to store generated files in this repo.

PR #338, opened 2024-03-08
chore: remove manifests folder

ksimon1:
chore: remove manifests folder

This commit removes manifests folder and adds it to .gitignore file, because this folder is not necessary to have permanently in the repository. However the scripts will still generate it, because it is helpfull during releases.

**Release note**:
```
NONE
```


0xFelix (approved):
/approve

PR #339, opened 2024-03-08
Switch git versioning

ksimon1:
Switch to git versioning

this PR changes all scripts necessary to generate task manifests.
Scripts will now generate task manifests into release/tasks folder
instead of tasks folder. At the same time this commit removed tag
creation from release/tasks/*/ folder, because we will be swithing to git based
versioning. This change will help reduce number of manifests located in
repository - at the end we will have only templates and generated
manifests at the repo.

**Release note**:
```
NONE
```


.github/workflows/artifact-hub-linter.yaml
```diff
@@ -1,12 +1,13 @@
-name: artifact hub linter
-on:
-  pull_request:
-    branches: [ main ]
+#Disable this job until https://github.com/artifacthub/hub/issues/3691 is resolved
```
0xFelix, line 1: Create an issue for it to be reenabled?

0xFelix (approved):
Looks good to me

/lgtm

.github/workflows/artifact-hub-linter.yaml
ksimon1: https://github.com/kubevirt/kubevirt-tekton-tasks/issues/341

PR #345, opened 2024-03-14
chore: bump version to v0.19.0

ksimon1:
chore: bump version to v0.19.0

This commit bumps all versions of image to v0.19.0. This is necessary, because artifacthub takes files from GH tagged release and in case the images would not have newer version before release, release would contain older version

After this PR is merged, I will create new release which will push new containers to quay
**Release note**:
```
NONE
```


0xFelix (approved):
/lgtm

PR #346, opened 2024-03-14
fix: fix PipelineRun examples

ksimon1:
fix: fix PipelineRun examples

this commit fixes wrong kind in pipelinerun and adds security context in efi pipelineruns. This is necessary when PipelineRun uses resolver. Then the modify-windows-iso-file task runs under different user and group.

**Release note**:
```
NONE
```

### DIFF
--- a/release/pipelines/windows-efi-installer/README.md
+++ b/release/pipelines/windows-efi-installer/README.md
@@ -78,12 +78,19 @@ spec:
         -   name: type
             value: artifact
         -   name: kind
-            value: task
+            value: pipeline
         -   name: name
             value: windows-efi-installer
         -   name: version
             value: v0.19.0
         resolver: hub
+    taskRunSpecs:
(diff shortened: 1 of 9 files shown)


0xFelix (approved):
/lgtm

ksimon1:
Merging even then pipeline tests fails. The failures are not related to this PR.

release/pipelines/windows-efi-installer/README.md
```diff
@@ -111,12 +118,19 @@ spec:
...
             value: artifact
         -   name: kind
-            value: task
+            value: pipeline
         -   name: name
             value: windows-efi-installer
         -   name: version
             value: v0.19.0
         resolver: hub
+    taskRunSpecs:
```
dominikholler, line 127: Thanks @ksimon1 for finding this issue!
Why is the explicit  securityContext setting required, and could this be set via the UI?

PR #347, opened 2024-03-14
chore: remove step which update versions in release

ksimon1:
chore: remove step which update versions in release

this commit removes the step, which updates manifests with new version. This step is not necessary, because the version has to be bumbed before release is created.

**Release note**:
```
NONE
```


0xFelix (approved):
/lgtm

PR #348, opened 2024-03-15
fix: url of getisourl.py script in pipeline readmes

ksimon1:
fix: url of getisourl.py script in pipeline readmes

**Release note**:
```
fix: url of getisourl.py script in pipeline readmes
```


jcanocan:
Thanks @ksimon1!
/lgtm

PR #357, opened 2024-03-20
chore: enable setting catalog version without "v" prefix

ksimon1:
chore: enable setting catalog version without "v" prefix
this commit allows to set REMOVE_CATALOG_VERSION_PREFIX env variable, which will remove "v" prefix from version in catalog version - e.g. 0.19.0 when true vs v0.19.0 when false.

This will help in DS where OCP catalog removes v from version. This commit also replaces all version variables from catalog version and replaces it with catalog_version which can hold version with or without v prefix

Signed-off-by: Karel Simon <ksimon@redhat.com>

**Release note**:
```
NONE
```

### DIFF
--- a/scripts/ansible/common.yaml
+++ b/scripts/ansible/common.yaml
@@ -13,6 +13,8 @@ sa_name: "{{ role_name }}"
 role_binding_name: "{{ role_name }}"
 default_file_mode: "0644"
 version: "{{ lookup('env','RELEASE_VERSION')| default('latest', true) }}"
+remove_catalog_version_prefix: "{{ lookup('env','REMOVE_CATALOG_VERSION_PREFIX')| default(false) | bool}}"
+catalog_version: "{{ version[1:] if remove_catalog_version_prefix else version }}"
 
 windows10: windows10
 windows11: windows11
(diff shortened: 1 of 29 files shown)


scripts/ansible/common.yaml
```diff
@@ -13,6 +13,8 @@ sa_name: "{{ role_name }}"
 role_binding_name: "{{ role_name }}"
 default_file_mode: "0644"
 version: "{{ lookup('env','RELEASE_VERSION')| default('latest', true) }}"
+remove_catalog_version_prefix: "{{ lookup('env','REMOVE_CATALOG_VERSION_PREFIX')| default(false, true) | bool}}"
```
0xFelix, line 16:
```suggestion
remove_catalog_version_prefix: "{{ lookup('env','REMOVE_CATALOG_VERSION_PREFIX')| default(false) | bool}}"
```

Shouldn't it look like this?

ksimon1: According to docs (https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#providing-default-values)
`If you want to use the default value when variables evaluate to false or an empty string you have to set the second parameter to true:`
So I assume the true can stay

0xFelix: I think we're both right in this specific case. When using bools we don't care for empty vars as they evaluate to false too. But in other cases your solution might be preferred!

PR #358, opened 2024-03-27
chore: remove default value from baseDVNamespace and sourceDiskImageNamespace

ksimon1:
chore: remove default value from baseDVNamespace and sourceDiskImageNamespace in pipelines

Tekton probably contains bug, which does not propagate params from PipelineRuns, when resolver is defined. Instead it uses default values defined in pipeline manifest -
https://issues.redhat.com/browse/SRVKP-4329.
Hovewer if the param does not contain any default value, tekton passes correct value

**Release note**:
```
chore: remove default value from baseDVNamespace and sourceDiskImageNamespace in pipelines
```

### DIFF
--- a/release/pipelines/windows-bios-installer/README.md
+++ b/release/pipelines/windows-bios-installer/README.md
@@ -42,6 +42,8 @@ The Pipeline implements this by spinning up a new VirtualMachine which boots fro
 
 Before you create PipelineRuns, you must create ConfigMaps with an autounattend.xml in the same namespace in which the VirtualMachine will be created.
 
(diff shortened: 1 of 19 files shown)


0xFelix:
Doesn't this mean all params with default values are broken?

ksimon1:
Not all, only for some unknown reasons the baseDvNamespace uses the default value eventhought the PipelineRun contains different value. 
On the another hand, this change makes sense, because if regular user would like to try these pipelines, the user have to change these parameters.
Regular users can't create DVs in kubevirt-os-image namespace. So from my point it is better to remove the default value and keep user to decide where the DVs should be.

ksimon1:
Closing, will be replaced by https://github.com/kubevirt/kubevirt-tekton-tasks/pull/360

PR #360, opened 2024-04-02
chore: remove namespace params from pipelines

ksimon1:
chore: remove namespace params from pipelines
Since the regular users doesn't normally have access to different namespaces, multi namespace use case doesn't make sense. And since we want to have the pipelines as simple as possible, the multi namespace use case adds a lot of complexity with creating RBAC objects and correctly binding them to pipelines.

**Release note**:
```
chore: remove namespace params from pipelines
```


0xFelix (approved):
Thanks!

/lgtm

PR #368, opened 2024-04-08
chore: improve logs in modify-data-object

ksimon1:
chore: improve logs in modify-data-object

this commit adds functionality, which will print out, when task is waiting for import to be finished, or when waiting until object is deleted.

**Release note**:
```
NONE
```

### DIFF
--- a/modules/modify-data-object/pkg/dataobject/provider.go
+++ b/modules/modify-data-object/pkg/dataobject/provider.go
@@ -72,6 +72,7 @@ func (d *dataObjectProvider) DeletePVC(namespace string, name string) error {
 }
 
 func waitUntilDataObjectIsDeleted(helper *resource.Helper, namespace, name string) {
+	log.Logger().Info("waiting until data object is deleted")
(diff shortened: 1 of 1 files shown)


jcanocan:
Thanks @ksimon1!
/lgtm

modules/modify-data-object/pkg/dataobject/provider.go
```diff
@@ -72,6 +72,7 @@ func (d *dataObjectProvider) DeletePVC(namespace string, name string) error {
 }
 
 func waitUntilDataObjectIsDeleted(helper *resource.Helper, namespace, name string) {
+	log.Logger().Info("waiting until Data object is deleted")
```
0xFelix, line 75: nit: Use same case in all messages? `data object` or `Data object`

ksimon1: fixed

0xFelix (approved):
/lgtm

Thanks!

PR #383, opened 2024-04-26
chore: download bootsources from quay instead of dockerhub

ksimon1:
chore: download bootsources from quay instead of dockerhub in e2e tests

increase memory requirements for VMs

**Release note**:
```
chore: download bootsources from quay instead of dockerhub in e2e tests
```


0xFelix (approved):
/lgtm

PR #384, opened 2024-04-26
fix: downgrade sigs.k8s.io/kustomize packages

ksimon1:
fix: downgrade sigs.k8s.io/kustomize packages
k8s.io/cli-runtime requires older versions of sigs.k8s.io/kustomize packages

**Release note**:
```
NONE

```


akrejcir:
Looks good. When we have time in the future, we could go through the dependencies and decide if these `replace` blocks are needed.

/lgtm

PR #392, opened 2024-05-03
fix: folders in gitignore

ksimon1:
fix: folders in gitignore
Fix rules to ignore folder manifests and .idea only in the root of the project

**Release note**:
```
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 **/dist/
-.idea/
+./.idea/
 ah
-manifests
-test/test.test
+/manifests/
+/test/test.test


.gitignore
```diff
@@ -1,5 +1,5 @@
 **/dist/
-.idea/
+./idea/
```
0xFelix, line 2:
```suggestion
./.idea/
```
Missing a dot?

ksimon1: correct :+1: , fixed

0xFelix (approved):
/lgtm

PR #404, opened 2024-05-17
fix: create-vm task if VMI manifest is used

ksimon1:
fix: create-vm task if VMI manifest is used

when VMI manifest does not contain spec.template.objectMeta, the task failed, due to nil pointer. This fix checks if objectMeta exists and if not, use VMI objectMeta instead

**Release note**:
```
fix: create-vm task if VMI manifest is used
```


jcanocan:
Thanks @ksimon1!
/lgtm

PR #410, opened 2024-05-23
fix: put false value in accept eula xml tag

ksimon1:
fix: put false value in accept eula xml tag

instead of commenting the eula, put a false value in accept eula tag.
update README
**Release note**:
```
fix: put false value in accept eula xml tag
```

### DIFF
--- a/release/pipelines/windows-efi-installer/configmaps/windows-efi-installer-configmaps.yaml
+++ b/release/pipelines/windows-efi-installer/configmaps/windows-efi-installer-configmaps.yaml
@@ -98,7 +98,7 @@ data:
             </OSImage>
           </ImageInstall>
           <UserData>
-            <!-- <AcceptEula>true</AcceptEula> -->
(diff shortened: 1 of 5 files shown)


release/pipelines/windows-efi-installer/configmaps/windows-efi-installer-configmaps.yaml
```diff
@@ -250,7 +250,7 @@ data:
                     </OSImage>
                 </ImageInstall>
                 <UserData>
-                    <AcceptEula>true</AcceptEula>
+               <!-- <AcceptEula>true</AcceptEula> -->
```
0xFelix, line 253: Would it make sense to set it to false instead of making it a comment?

ksimon1: Both ways will work :) Your approach makes more sense

0xFelix (approved):
/lgtm
/approve

PR #429, opened 2024-06-13
chore: change the IDs in securityContext of PipelineRuns/TaskRuns

ksimon1:
chore: change the IDs in securityContext of PipelineRuns/TaskRuns

The IDs defined in securityContext of example PipelineRuns are not accurate. The disk has preset IDs 107 (qemu user). It makes sense to use the same IDs for PipelineRun / TaskRun.
Added mention and solution about issue with permission denied in disk-virt tasks readmes.

**Release note**:
```
NONE
```


jcanocan:
This looks great! Thanks!
/lgtm

PR #437, opened 2024-06-21
fix: readme and efi pipeline description

ksimon1:
fix: readme and efi pipeline description
add missing windows 10 support to description of efi pipeline and README

**Release note**:
```
NONE
```


0xFelix (approved):
/lgtm

PR #439, opened 2024-06-25
chore: add useBiosMode parameter to efi-pipeline

ksimon1:
chore: add useBiosMode parameter to efi-pipeline

Windows preferences in older KubeVirt versions might still use Bios mode. In that case, set the `useBiosMode` parameter to `true`. This will skip the `modify-windows-iso-file` task. In case the the Windows preference uses Bios and the `useBiosMode` parameter is not set to true, the Windows VM will not work.

**Release note**:
```
chore: add useBiosMode parameter to efi-pipeline
```

### DIFF
--- a/release/pipelines/windows-efi-installer/README.md
+++ b/release/pipelines/windows-efi-installer/README.md
@@ -235,3 +235,6 @@ spec:
           fsGroup: 107
           runAsUser: 107
 ```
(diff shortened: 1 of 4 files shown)


0xFelix (commented):
So essentially this is now a universal pipeline for EFI and BIOS?

ksimon1:
> So essentially this is now a universal pipeline for EFI and BIOS?

Correct, but I am investigate the possibility to don't do it this way, but set correct params to xorriso and do hybrid iso.

PR #447, opened 2024-07-03
chore: log hash of the iso everytime

ksimon1:
chore: log hash of the iso everytime

modify-windows-iso-file task should log the iso hash everytime not only during error

**Release note**:
```
chore: log hash of the iso everytime
```


jcanocan:
Thanks for the improvement!
/lgtm

PR #451, opened 2024-07-09
fix: unit tests

ksimon1:
fix: unit tests

this commit is fixing triggering of unit tests

**Release note**:
```
NONE
```

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,24 +1,7 @@
 #!/usr/bin/env bash
 
-SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
-REPO_DIR="$(realpath "${SCRIPT_DIR}/..")"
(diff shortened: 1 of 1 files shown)


scripts/test.sh
```diff
@@ -1,24 +1,7 @@
 #!/usr/bin/env bash
 
-SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
-REPO_DIR="$(realpath "${SCRIPT_DIR}/..")"
+cd modules ||exit 1
```
0xFelix, line 3:
```suggestion
cd modules || exit 1
```

nit

ksimon1: fixed :)

0xFelix (approved):
/lgtm

Thanks!

PR #459, opened 2024-07-19
chore: Bump dependencies and go version

ksimon1:
chore: Bump dependencies and go version

This PR is based on https://github.com/kubevirt/kubevirt-tekton-tasks/pull/458 and fixes the code

**Release note**:
```
chore: Bump dependencies and go version
```


go.mod
0xFelix, line 1: Are most of these replaces still necessary?

ksimon1: These replaces are mainly because KubeVirt was using old k8s dependencies. If KubeVirt project will update the k8s dependencies regularly, then they are not required

0xFelix: In vm-console-proxy many replaces could be dropped. Can you please check which ones are still required?

See https://github.com/kubevirt/vm-console-proxy/pull/159

ksimon1: The required one is only the 
```
replace k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20240430033511-f0e62f92d13f
```
But again, eventhough Kubevirt bumped now the dependencies, it doesn't mean we will do it periodically (eventhough we should :) ) and there is a chance we will have to introduce this block in the future again.

0xFelix: Not really a fan of keeping things just in case but ultimately it is your call.

ksimon1: I removed all unnecessary replace statements :+1:

akrejcir (commented):
/lgtm

PR #467, opened 2024-07-23
chore: enable renovate for older release branches

ksimon1:
enable renovate for these branches: release-v0.15, release-v0.13, release-v0.12

**Release note**:
```
NONE
```


akrejcir:
So the configuration file that specifies which branches should be checked is in the `main` branch? That is strange convention.

/lgtm

ksimon1:
Yes, but that gives you flexibility to decouple the settings for each branch and have simple config for each branch, instead of single complex config

PR #489, opened 2024-07-29
chore: update config of renovate

ksimon1:
chore: update config of renovate

this commit adds ignorePaths to ignore modules/tests folder, because it updates a lot of packages which are unnecessary in release branches.
Removed constraints with hardcoded go version for main. 
Moved extends to the top of the file.

**Release note**:
```
NONE
```

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,7 @@
 {
+  "extends": [
+    ":gitSignOff"
+  ],
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "baseBranches": [
     "main",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -19,8 +19,8 @@
   "labels": [
     "release-note-none"
   ],
-  "extends": [
-    ":gitSignOff"
+  "ignorePaths": [
+    "modules/tests/**"
```
akrejcir, line 23: I don't see the `modules/tests` directory. Is it generated and not comitted to the repo?

ksimon1: This path modules/tests/ exists only in releases 4.15 and older. The 4.16 is using different project structure (and this rule is not affecting anything).

akrejcir: ok

PR #497, opened 2024-07-30
chore: add go package to version control for release branches

ksimon1:
chore: add go package to version control for release branches

Add go package to packageRules so we are able to control which go version is in which branch

**Release note**:
```
NONE

```


lmilbaum:
LGTM

PR #541, opened 2024-10-31
fix: hardcode instance-types version to 1.1.0

ksimon1:
fix: hardcode instance-types version to 1.1.0

newest instance types (v1.2.0) can't work with kubevirt 1.3.1. We have to wait until the new kubevirt version is out

**Release note**:
```
NONE
```

### DIFF
--- a/automation/e2e-deploy-resources.sh
+++ b/automation/e2e-deploy-resources.sh
@@ -18,8 +18,10 @@ TEKTON_VERSION=$(curl -s https://api.github.com/repos/tektoncd/operator/releases
 SSP_OPERATOR_VERSION=$(curl -s  https://api.github.com/repos/kubevirt/ssp-operator/releases | \
             jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
 
-INSTANCE_TYPES_VERSION=$(curl -s  https://api.github.com/repos/kubevirt/common-instancetypes/releases | \
-            jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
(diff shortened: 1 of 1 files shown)


automation/e2e-deploy-resources.sh
```diff
@@ -18,8 +18,10 @@ TEKTON_VERSION=$(curl -s https://api.github.com/repos/tektoncd/operator/releases
 SSP_OPERATOR_VERSION=$(curl -s  https://api.github.com/repos/kubevirt/ssp-operator/releases | \
             jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
 
-INSTANCE_TYPES_VERSION=$(curl -s  https://api.github.com/repos/kubevirt/common-instancetypes/releases | \
-            jq '.[] | select(.prerelease==false) | .tag_name' | sort -V | tail -n1 | tr -d '"')
+INSTANCE_TYPES_VERSION="v1.1.0"
```
lyarwood, line 21: Please go ahead if you need this to unblock something but these tests should just be using the KubeVirt FG with <= v1.3.0 to deploy common instancetypes. That FG is now GA'd in v1.4.0 and as such is enabled by default so you don't even need enable it.

ksimon1: Would you please point me to the name of the feature gate which deploys instance types in Kubevirt CR?

jcanocan: The feature gate is `CommonInstancetypesDeploymentGate`, but after v1.4.0, the FG has been replaced with a configurable (enabled by default). See: https://github.com/kubevirt/kubevirt/pull/12753

PR #543, opened 2024-11-03
feat(CNV-40411): set default instancetype and preference labels

codingben:

During execution of the disk-uploader, it should
be able to get the default-instancetype and
default-preference labels from the DV or PVC,
and set them in the newly created containerDisk
metadata afterwards.

**Release note**:

```release-note
Set instancetype and preference on a newly created containerDisk before uploading it
```


modules/disk-uploader/pkg/image/image.go
```diff
@@ -16,7 +17,24 @@ import (
...
-func Build(diskPath string) (v1.Image, error) {
+const (
+	defaultInstanceType = "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_INSTANCETYPE"
+	defaultPreference   = "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_PREFERENCE"
+)
+
+func DefaultConfig(labels map[string]string) v1.Config {
+	var env []string
+
+	if value, exists := labels["instancetype.kubevirt.io/default-instancetype"]; exists {
```
ksimon1, line 28: please use constant for `instancetype.kubevirt.io/default-instancetype`, the same for preference

modules/disk-uploader/pkg/image/image.go
```diff
@@ -16,7 +17,24 @@ import (
 	tar "kubevirt.io/containerdisks/pkg/build"
 )
 
-func Build(diskPath string) (v1.Image, error) {
+const (
+	defaultInstanceType = "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_INSTANCETYPE"
+	defaultPreference   = "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_PREFERENCE"
+)
+
+func DefaultConfig(labels map[string]string) v1.Config {
```
codingben, line 25: @0xFelix @lyarwood Currently this function returns a config with default instancetype and preference, what would we do in case of custom instancetype or preference?

lyarwood: @codingben custom in terms of something different being used by a `VirtualMachine` or `VirtualMachineSnapshot`? We could just override the defaults with the values used there but I'm not sure I entirely understand the flow here tbh.

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -99,6 +99,43 @@ func GetRawDiskUrlFromVolumes(client kubecli.KubevirtClient, namespace, name, vo
...
+		}
+		sourceLabels = pvc.GetLabels()
+	default:
+		return nil, fmt.Errorf("unsupported source kind: %s", source.Kind)
+	}
+
+	return filterLabels(sourceLabels), nil
+}
+
+func filterLabels(labels map[string]string) map[string]string {
```
lyarwood, line 129: This isn't required.

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -99,6 +99,43 @@ func GetRawDiskUrlFromVolumes(client kubecli.KubevirtClient, namespace, name, vo
...
 }
 
+func GetLabelsFromExportSource(k8sClient *k8sv1.CoreV1Client, virtClient kubecli.KubevirtClient, exportSourceKind, exportSourceNamespace, exportSourceName, volumeName string) (map[string]string, error) {
+	source, err := getExportSource(exportSourceKind, exportSourceName)
+	if err != nil {
+		return nil, err
+	}
+
+	var sourceLabels map[string]string
+	switch source.Kind {
```
lyarwood, line 109: This duplicates the switch in `getExportSource`, I'd remove the call to `getExportSource` and just switch on `exportSourceKind` here.

lyarwood (changes requested):

modules/disk-uploader/pkg/image/image.go
0xFelix, line 25: I'd suggest to use regexes/replacements here. One to filter the labels we want, and another one to convert the label name to env variable.

See for matching regex:
https://github.com/kubevirt/containerized-data-importer/blob/1f34e0ea9d5d5baa00478c4c466f3c451e84c557/pkg/controller/common/util.go#L386

Create inverse of this:
https://github.com/kubevirt/containerized-data-importer/blob/1f34e0ea9d5d5baa00478c4c466f3c451e84c557/pkg/importer/util.go#L77

codingben: @lyarwood Can be there something else than `instancetype.kubevirt.io/default-instancetype` or `instancetype.kubevirt.io/default-preference`? not a default one?

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: Why not? I want to make sure that it returns only required labels and not all the labels.

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: Yep!

modules/disk-uploader/pkg/image/image.go
0xFelix: I don't think it is more or less complicated than adding a `filterLabels` func.

ksimon1: Is the https://github.com/kubevirt/containerized-data-importer/blob/1f34e0ea9d5d5baa00478c4c466f3c451e84c557/pkg/importer/util.go#L77 function used for all labels/annotations?

lyarwood: We also support `-kind` variants of both of these labels:

https://github.com/kubevirt/kubevirt/blob/0436bc52bf02c689190cc86889c3ec90b8bcfecf/staging/src/kubevirt.io/api/instancetype/register.go#L42-L47

modules/disk-uploader/pkg/vmexport/vmexport.go
lyarwood: Seems a little redundant to filter when the code extracting the label values within `DefaultConfig` would work without this?

modules/disk-uploader/pkg/image/image.go
0xFelix: `envToLabel` is used in CDI when importing from a registry source. It only converts labels.

codingben: > I don't think it is more or less complicated than adding a filterLabels func.

The `filterLabels` function was removed. Thanks for sharing CDI functions, it's useful to get famliar with them in that scenario.

Can you please help me to understand what labels needed except the 2 mentioned here? I don't think we should set more labels than what we need in the image config below:

```
func DefaultConfig(labels map[string]string) v1.Config {
	var env []string

	if value, exists := labels[defaultInstanceTypeLabel]; exists {
		env = append(env, fmt.Sprintf("%s=%s", defaultInstanceTypeEnv, value))
	}
	if value, exists := labels[defaultPreferenceLabel]; exists {
		env = append(env, fmt.Sprintf("%s=%s", defaultPreferenceEnv, value))
	}
	return v1.Config{Env: env}
}
```

0xFelix: As Lee mentioned, it should at least also look after the `-kind` variants of these labels.

codingben: Yes I saw it. I've modified it in this [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/fdc80d265fe74f103a1ce3fe8e2442fc3366b58a..a3b9ff76d4161b586c65313e23d4743577f1829e). Can you please take a look?

codingben: @lyarwood Thanks. I've modified it in this last [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/fdc80d265fe74f103a1ce3fe8e2442fc3366b58a..a3b9ff76d4161b586c65313e23d4743577f1829e).

codingben:
@ksimon1 @0xFelix @lyarwood Can you please review again? :)

modules/disk-uploader/pkg/image/image.go
```diff
@@ -16,7 +17,26 @@ import (
 	tar "kubevirt.io/containerdisks/pkg/build"
 )
 
-func Build(diskPath string) (v1.Image, error) {
+var labelEnvPairs = []struct {
+	label, env string
+}{
+	{"instancetype.kubevirt.io/default-instancetype", "INSTANCETYPE_KUBEVIRT_IO_DEFAULT_INSTANCETYPE"},
```
0xFelix, line 23: I'd use the simpler map[string]string and would iterate over it down below.

codingben: Yep that would be better, but I don't see it as key:value here, IMO both values are keys (key:key).

codingben:
@0xFelix This [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/a3b9ff76d4161b586c65313e23d4743577f1829e..a35cdca7549ab660fdd22e7ae3039964cf0ca18c) will apply key:value that you requested in your latest review. Can you please take a look?

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -99,6 +99,28 @@ func GetRawDiskUrlFromVolumes(client kubecli.KubevirtClient, namespace, name, vo
 	return "", fmt.Errorf("volume %s is not found in VirtualMachineExport internal volumes", volumeName)
 }
 
+func GetLabelsFromExportSource(k8sClient *k8sv1.CoreV1Client, virtClient kubecli.KubevirtClient, exportSourceKind, exportSourceNamespace, exportSourceName, volumeName string) (map[string]string, error) {
+	var sourceLabels map[string]string
+	switch exportSourceKind {
+	case "VirtualMachine", "VirtualMachineSnapshot":
+		dv, err := virtClient.CdiClient().CdiV1beta1().DataVolumes(exportSourceNamespace).Get(context.Background(), volumeName, metav1.GetOptions{})
```
0xFelix, line 106: Can you explain why VM/VMS equals DV? What if the volume of a VM is a PVC without DV?

codingben: From what I see, `datavolumedisk` is what set in the VM snapshot - From the [example-vm](https://github.com/kubevirt/kubevirt-tekton-tasks/blob/main/examples/pipelines/disk-uploader/disk-uploader-pipeline.yaml#L31-L77) volumes in case you apply it:
```
...
volumes:
   - dataVolume:
     name: example-dv
   name: datavolumedisk
```
I thought that in both kinds, if you specify `example-dv` volume name it'll work. I'll need to check it and test it (haven't tested it yet, because wanted you to review the IT/Pref labels).

codingben:
I need to test this PR changes, and add unit tests after it. I'll mark it as draft at the moment.

modules/disk-uploader/pkg/image/image.go
codingben: @0xFelix @lyarwood Are you okay with current implementation? Can we resolve this discussion?

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: > I thought that in both kinds, if you specify example-dv volume name it'll work.

I can see that if you apply VMExport on both VM and VMSnapshot, it provides the same volume name (`example-dv` as I mentioned). So this case: `case "VirtualMachine", "VirtualMachineSnapshot"` is valid.

> What if the volume of a VM is a PVC without DV?

In that case, there is a single volume that is name is the PVC name.

codingben:
This [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/d0f3d35e4f23aec446bd5810b7d05efe9a751135..225f69c323b79748361dd466cb4092de355fb13f) is a rebase to have unit tests in this PR.

codingben:
This [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/225f69c323b79748361dd466cb4092de355fb13f..89ce942291805fc4b6366735db7f64d641269b2d) is have unit tests to test the `GetLabelsFromExportSource()` function from `vmexport.go`.

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: I've added here [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/225f69c323b79748361dd466cb4092de355fb13f..89ce942291805fc4b6366735db7f64d641269b2d) to test the `GetLabelsFromExportSource()` function from `vmexport.go`.

0xFelix: I'd suggest to use an if statement for the VM/Snapshot case and let it fall through to PVC handling, if no DV was found on the VM/Snapshot, instead of duplicating the PVC handling code.

modules/disk-uploader/pkg/vmexport/vmexport_test.go
```diff
@@ -186,4 +191,68 @@ var _ = Describe("VMExport", func() {
...
+					Namespace: namespace,
+					Labels: map[string]string{
+						"instancetype.kubevirt.io/default-instancetype": "cx1.2xlarge",
+						"instancetype.kubevirt.io/default-preference":   "fedora",
+					},
+				},
+			}, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			labels, err := vmexport.GetLabelsFromExportSource(virtClient, "VirtualMachine", namespace, name, "example-volume")
```
0xFelix, line 209: Make it a table for "VirtualMachine" and "VirtualMachineSnapshot"? Add test for all cases where no DV/PVC could be found?

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: duplicating the PVC call is not an option, because you need to provide different names, in case of VM or VMSnapshot kinds, PVC name equals to volume name:
```
virtClient.CoreV1().PersistentVolumeClaims(exportSourceNamespace).Get(context.Background(), volumeName, metav1.GetOptions{})
```
And in case of PVC kind, then just the PVC name:
```
virtClient.CoreV1().PersistentVolumeClaims(exportSourceNamespace).Get(context.Background(), exportSourceName, metav1.GetOptions{})
```

codingben: Also IMO, switch cases are more testable, clearer, and readable. It makes sense to think of it like if VM or VMSnapshot is not found, then try to get it from PVC directly.

modules/disk-uploader/pkg/vmexport/vmexport_test.go
codingben: Done.

modules/disk-uploader/pkg/vmexport/vmexport.go
0xFelix: You've added a lot of nesting into the switch case, I don't think that makes it more readable. If you want to keep it separate I'd suggest to add two funcs for the different cases and to call them from the switch cases.

modules/disk-uploader/pkg/vmexport/vmexport_test.go
```diff
@@ -186,4 +191,84 @@ var _ = Describe("VMExport", func() {
...
+			Entry("for VirtualMachineSnapshot", "VirtualMachineSnapshot"),
+		)
+
+		DescribeTable("should return labels from pvc if there is no datavolume", func(resourceType string) {
+			_, err := kubeClient.CoreV1().PersistentVolumeClaims(namespace).Create(context.Background(), &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "example-pvc",
+					Namespace: namespace,
+					Labels: map[string]string{
+						"instancetype.kubevirt.io/default-instancetype": "cx1.2xlarge",
```
0xFelix, line 224: You could add consts for these.

modules/disk-uploader/pkg/vmexport/vmexport_test.go
```diff
@@ -186,4 +191,84 @@ var _ = Describe("VMExport", func() {
...
+		})
+
+		It("should return unsupported source kind error", func() {
+			_, err := vmexport.GetLabelsFromExportSource(virtClient, "VirtualMachineInstance", namespace, name, "disk-volume-1")
+			Expect(err).To(MatchError("unsupported source kind: VirtualMachineInstance"))
+		})
+
+		DescribeTable("should return an error when DV or PVC is not found", func(exportSourceKind string, volumeName string) {
+			labels, err := vmexport.GetLabelsFromExportSource(virtClient, exportSourceKind, namespace, name, volumeName)
+			Expect(err).To(HaveOccurred())
```
0xFelix, line 266: Assert the specific error?

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: Done.

modules/disk-uploader/pkg/vmexport/vmexport_test.go
```diff
@@ -186,4 +191,89 @@ var _ = Describe("VMExport", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			labels, err := vmexport.GetLabelsFromExportSource(virtClient, "PersistentVolumeClaim", namespace, "example-pvc", "")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(labels).To(HaveKeyWithValue(defaultInstanceType, "cx1.2xlarge"))
+			Expect(labels).To(HaveKeyWithValue(defaultPreference, "fedora"))
+		})
+
+		It("should return unsupported source kind error", func() {
+			_, err := vmexport.GetLabelsFromExportSource(virtClient, "VirtualMachineInstance", namespace, name, "disk-volume-1")
```
0xFelix, line 265: nit: Add nil check as in test case below

0xFelix (commented):
Looking great, just one more nit!

codingben:
@0xFelix Resolved your last review in this [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/84ff8086fba20def2ce3dace55f6ba382f962a91..d0999c6835525040193125d314c6b856f6b71a23).

0xFelix (approved):
Thanks

ksimon1:
failing tests are not related to this change.

PR #560, opened 2024-11-27
fix(CNV-52011): create objects with random names

### LINKED ISSUE #542: disk uploader task panics when secret already exists
**What happened**:
When user defines in Pipeline retries attribute for disk uploader task as e.g. example: 
```
    - name: upload-image-to-registries
      timeout: 9h0m0s
      retries: 2
```
the uploader task during second run panics with:

```
W1101 10:28:41.020869 18 client_config.go:659] Neither --kubeconfig nor --master was specified. Using the inClusterConfig. This might not work.
{"level":"info","ts":1730456921.0233357,"caller":"disk-uploader/main.go:39","msg":"Creating a new Secret object...","namespace":"windows-builder","name":"windows-efi-2nmxw"}
panic: secrets "windows-efi-2nmxw" already exists [recovered]
panic: secrets "windows-efi-2nmxw" already exists
goroutine 1 [running]:
github.com/kubevirt/kubevirt-tekton-tasks/modules/shared/pkg/exit.HandleExit()
/src/modules/shared/pkg/exit/exit.go:80 +0xbe
panic({0x1c7b420?, 0xc0006109e0?})
/root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.6.linux-amd64/src/runtime/panic.go:770 +0x132
github.com/kubevirt/kubevirt-tekton-tasks/modules/shared/pkg/exit.ExitOrDieFromError(0x3, {0x20c3c00, 0xc0003fd360}, {0x0, 0x0, 0x11?})
/src/modules/shared/pkg/exit/exit.go:60 +0x134
main.main()
/src/cmd/disk-uploader/main.go:135 +0x385
```
**What you expected to happen**:
The task generates secret with random name

**How to reproduce it (as minimally and precisely as possible)**:
100%

/cc @codingben

codingben:

When running disk-uploader it creates
VMExport object and VMExport secret
to store a token that required by the
Export API endpoints.

Currently the name of VMExport object
and VMExport secret is not random
generated and will fail to be created
again if the disk-uploader (as a Tekton
Task) will run twice. As a result, it
throws a panic error that causes the
failure of the execution.

Fixes #542.

**Release note**:

```release-note
Generate random name of objects
```

### DIFF
--- a/modules/disk-uploader/pkg/secrets/secrets.go
+++ b/modules/disk-uploader/pkg/secrets/secrets.go
@@ -13,17 +13,17 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func CreateVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace, name string) error {
+func CreateVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace, baseName string) (*corev1.Secret, error) {
 	length := 20
 	token, err := generateSecureRandomString(length)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	v1Secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: namespace,
+			GenerateName: baseName + "-",
+			Namespace:    namespace,
 		},
 		StringData: map[string]string{
 			"token": token,
@@ -31,11 +31,10 @@ func CreateVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace,
 	}
 
 	if err := ownerreference.SetPodOwnerReference(k8sClient, v1Secret); err != nil {
-		return err
+		return nil, err
 	}
 
-	_, err = k8sClient.CoreV1().Secrets(namespace).Create(context.Background(), v1Secret, metav1.CreateOptions{})
-	return err
+	return k8sClient.CoreV1().Secrets(namespace).Create(context.Background(), v1Secret, metav1.CreateOptions{})
 }
 
 func GetTokenFromVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace, name string) (string, error) {
--- a/cmd/disk-uploader/main.go
+++ b/cmd/disk-uploader/main.go
@@ -40,32 +40,34 @@ func run(opts parse.CLIOptions, k8sClient kubernetes.Interface, virtClient kubec
 
 	log.Logger().Info("Creating a new Secret object...", zap.String("namespace", namespace), zap.String("name", name))
 
-	if err := secrets.CreateVirtualMachineExportSecret(k8sClient, namespace, name); err != nil {
+	vmExportSecret, err := secrets.CreateVirtualMachineExportSecret(k8sClient, namespace, name)
+	if err != nil {
 		return err
 	}
 
 	log.Logger().Info("Creating a new VirtualMachineExport object...", zap.String("namespace", namespace), zap.String("name", name))
 
-	if err := vmexport.CreateVirtualMachineExport(virtClient, kind, namespace, name); err != nil {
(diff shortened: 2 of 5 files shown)


codingben:
Adding more reviewers because @ksimon1 is not available today.

/cc @lyarwood @akrejcir

modules/disk-uploader/pkg/secrets/secrets.go
```diff
@@ -22,7 +24,7 @@ func CreateVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace,
 
 	v1Secret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
+			Name:      generateRandomName(name),
```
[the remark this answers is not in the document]
codingben: I aimed to have similar naming to `generateSecureRandomString` above. I don't mind naming it `generateName`, but why it should be public function?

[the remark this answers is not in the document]
codingben: I renamed it to `generateName` in this [revision](https://github.com/kubevirt/kubevirt-tekton-tasks/compare/b78b6e03d0d98671a3773378e389824d1f732f98..9b539e15bb30d47802af7f29e4ac29396b5e8fda).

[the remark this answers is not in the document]
0xFelix: I meant, why are we not using the `GenerateName` feature offered by K8S? The field is named `GenerateName` instead of name and will append a random suffix.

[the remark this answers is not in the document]
codingben: Do you mean to `GenerateName` that is part of ObjectMeta? For example:
```
v1Secret := &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			GenerateName: name + "-",
...
}
```

codingben:
It also requires to generate the name of the VMExport, basically all the deployed Kubernetes objects by the disk-uploader.

codingben:
/unhold

@ksimon1 @0xFelix I've modified the PR, can you please review it again?

modules/disk-uploader/pkg/secrets/secrets.go
```diff
@@ -13,29 +13,28 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func CreateVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace, name string) error {
+func CreateVirtualMachineExportSecret(k8sClient kubernetes.Interface, namespace, name string) (*corev1.Secret, error) {
```
0xFelix, line 16: What about letting this func take care of the naming completely instead of having the `name` argument? It might come as a surprise to callers that the name of the created object is different than what was passed in.

codingben: Do you mean to have something like this: `GenerateName: "vmexport-secret-"`?

ksimon1: I would vote for having at least some connection to the original source, because then you can easily identify which object can be checked or deleted, if eg. something gets wrong

0xFelix: To identify created objects to delete you can take a look at the returned Secret, no?

0xFelix: I'd consider modifying the name a surprising side effect of this method at least. If you want to keep this then I'd suggest to rename the argument to `baseName` or something.

ksimon1: In the best case user should not care about these tmp objects, which are created just to achieve the upload of the disk. 
I agree the name param should be rename to something else.

codingben: Done.

codingben:
@ksimon1 @0xFelix Can you please review it again?

0xFelix (approved):
/lgtm

PR #562, opened 2024-11-28
fix(CNV-40411): add new disk-uploader roles

codingben:

The disk-uploader should be able to
access Data Volume or Presistent
Volume Claim to get the InstanceType
and Preference labels from them,
and set in the newly created
containerDisk.

**Release note**:

```release-note
None
```


0xFelix (approved):
/lgtm

PR #565, opened 2024-12-03
chore: return string instead of image interface in disk-uploader

ksimon1:
chore: return string instead of image interface in disk-uploader

Previously, the disk uploader run function returned image interface. It is not necesary, since we use only digest from it. That way, we get rid of image dependency in main.go file

**Release note**:
```
NONE
```


jcanocan:
Thanks! Nice idea.
/lgtm

PR #574, opened 2024-12-16
fix(CNV-53190): use emptyDir to store downloaded disk

codingben:

Currently the downloaded disk image from
the source will be stored inside container's
temporary file system, that consumes RAM,
and instead it should be stored in the node's
storage by using emptyDir.

**Release note**:

```release-note
Use emptyDir to store downloaded disk in disk-uploader.
```

### DIFF
--- a/release/tasks/disk-uploader/disk-uploader.yaml
+++ b/release/tasks/disk-uploader/disk-uploader.yaml
@@ -83,3 +83,9 @@ spec:
         memory: "3Gi"
       limits:
         memory: "5Gi"
+    volumeMounts:
+      - mountPath: /tmp
+        name: disk
+  volumes:
+    - name: disk
+      emptyDir: {}
--- a/templates/disk-uploader/manifests/disk-uploader.yaml
+++ b/templates/disk-uploader/manifests/disk-uploader.yaml
@@ -83,3 +83,9 @@ spec:
         memory: "3Gi"
       limits:
         memory: "5Gi"
+    volumeMounts:
+      - mountPath: /tmp
+        name: disk
+  volumes:
+    - name: disk
+      emptyDir: {}


0xFelix (commented):
Did you check what is a typical size for local ephemeral storage on nodes? It might be an issue to expect this storage to hold several hundred gigabytes.

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage

Alos, I think it's worth mentioning that streaming IO is preferable if possible.

codingben:
> Did you check what is a typical size for local ephemeral storage on nodes? It might be an issue to expect this storage to hold several hundred gigabytes.
> 
> https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage

Good catch Felix. I didn't check the typical size. I think from customer's perspective, it's cheaper to use storage than memory (it's also mentioned in the docs you shared).

We can set something like this, to guarantee the usage of disk-uploader:
```
...
    resources:
      requests:
        ephemeral-storage: "2Gi"
      limits:
        ephemeral-storage: "10Gi"
```
Altough I need to investigate the maximum disk size we can download from the source, rebuild as a containerDisk and upload to the container registry - I'm not sure about the values yet.

> Alos, I think it's worth mentioning that streaming IO is preferable if possible.

I agree, but streaming IO is not supported in Export API.

0xFelix:
> We can set something like this, to guarantee the usage of disk-uploader

Not sure what to recommend here. A disk might be larger than 10Gi before compressing it? How much scratch space is needed when working with a 10Gi disk? Will there be multiple copies of it?

dominikholler:
I expect that the resources limits and requests has to be customizable. I would postpone this adjustment after we understood the limitations of the default config of quay.io .

dominikholler:
My personal reference how to start is https://github.com/openshift-pipelines/tektoncd-catalog/blob/650a046c582e5a70719669fb7032a224ab42357e/tasks/task-buildah/0.5.0/task-buildah.yaml#L171 , because if it works for buildah, it would be a good starting point for the disk uploader task, too.

codingben:
> I expect that the resources limits and requests has to be customizable. I would postpone this adjustment after we understood the limitations of the default config of quay.io .

The image's storage config is set by `MAXIMUM_LAYER_SIZE` as mentioned in the Quay's docs [1]. The default config is 20gb and the user can set another value in the deployment of Quay.

[1] https://docs.projectquay.io/config_quay.html

PR #576, opened 2024-12-16
docs(CNV-53188): improve disk-uploader usage guidance

codingben:

Add more details to explain an usage of
EXPORT_SOURCE_KIND. Also, Add an example of the
Secret object, that will be referenced by
SECRET_NAME.

**Release note**:

```release-note
None
```

### DIFF
--- a/release/tasks/disk-uploader/README.md
+++ b/release/tasks/disk-uploader/README.md
@@ -12,14 +12,29 @@ When user runs [KubeVirt Tekton Tasks](https://github.com/kubevirt/kubevirt-tekt
 
 ### Parameters
 
-- **EXPORT_SOURCE_KIND**: The name of the export source kind
+- **EXPORT_SOURCE_KIND**: Specify the export source kind (Expected values `vm`, `vmsnapshot`, or `pvc`)
 - **EXPORT_SOURCE_NAME**: The name of the export source
 - **VOLUME_NAME**: The volume name (If source kind is PVC, then volume name is equal to source name)
 - **IMAGE_DESTINATION**: Destination of the image in container registry
-- **PUSH_TIMEOUT**: ContainerDisk push timeout in minutes
+- **PUSH_TIMEOUT**: ContainerDisk push timeout in minutes (Optional)
 - **SECRET_NAME**: Name of the secret which holds credential for container registry
 
 ### Usage
+
+To get the Secret of the task run:
+```
+apiVersion: v1
+data:
+    accessKeyId: <ACCESS_KEY_ID>
+    secretKey: <SECRET_KEY>
+kind: Secret
+metadata:
+    name: disk-uploader-credentials
+type: Opaque
+
+```
+
+Get `ACCESS_KEY_ID` or `SECRET_KEY` by running: `echo -n "<REGISTRY_USERNAME_OR_PASSWORD>" | base64`.
 
 Task run using resolver:
 ```
@@ -37,6 +52,8 @@ spec:
         value: example-dv
     -   name: IMAGE_DESTINATION
         value: quay.io/kubevirt/example-vm-exported:latest
+    -   name: SECRET_NAME
+        value: disk-uploader-credentials
     taskRef:
         params:
         -   name: catalog
--- a/release/tasks/disk-uploader/disk-uploader.yaml
+++ b/release/tasks/disk-uploader/disk-uploader.yaml
@@ -24,7 +24,7 @@ spec:
     to be used in multiple Kubernetes clusters.
   params:
   - name: EXPORT_SOURCE_KIND
-    description: The name of the export source kind
+    description: Specify the export source kind (Expected values "vm", "vmsnapshot", or "pvc")
     type: string
   - name: EXPORT_SOURCE_NAME
     description: The name of the export source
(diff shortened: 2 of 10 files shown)


release/tasks/disk-uploader/README.md
```diff
@@ -21,6 +21,19 @@ When user runs [KubeVirt Tekton Tasks](https://github.com/kubevirt/kubevirt-tekt
 
 ### Usage
 
+Secret of the task run:
+```
+apiVersion: v1
+data:
+    accessKeyId: <ACCESS_KEY_ID>
+    secretKey: <SECRET_KEY>
```
dominikholler, line 29: This requires still instructions how to get the ACCESS_KEY_ID and SECRET_KEY. E.g. the base64 encoding, and the command line to generate this comfortably. Maybe you can also mention that the instructions from https://tekton.dev/docs/pipelines/auth/#configuring-docker-authentication-for-docker will not work here.

dominikholler: Can you please point me to any documentation about the format of the accessKeyId and secretKey ? I think even CNV downstream doc is still missing it. If the task would accept the credentials in the default dockerconfigjson format like shown in quay.io's web ui or in tekton documented, I would agree with you. But this task is expecting a format which seems to be special for kubevirt, and so we should document it.

codingben: I've added an instruction how to get `ACCESS_KEY_ID` or `SECRET_KEY`.

release/tasks/disk-uploader/README.md
```diff
@@ -37,6 +52,10 @@ spec:
         value: example-dv
     -   name: IMAGE_DESTINATION
         value: quay.io/kubevirt/example-vm-exported:latest
+    -   name: PUSH_TIMEOUT
```
dominikholler, line 55: If the  PUSH_TIMEOUT would not be included in the example, the example would be more simple.

release/tasks/disk-uploader/README.md
```diff
@@ -12,7 +12,7 @@ When user runs [KubeVirt Tekton Tasks](https://github.com/kubevirt/kubevirt-tekt
 
 ### Parameters
 
-- **EXPORT_SOURCE_KIND**: The name of the export source kind
+- **EXPORT_SOURCE_KIND**: Specify the export source kind (vm, vmsnapshot, pvc)
```
jcanocan, line 15:
```suggestion
- **EXPORT_SOURCE_KIND**: Specify the export source kind (VM, VMsnapshot, PVC)
```
Please, use capitals to indicate that they are acronyms.

release/tasks/disk-uploader/disk-uploader.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     to be used in multiple Kubernetes clusters.
   params:
   - name: EXPORT_SOURCE_KIND
-    description: The name of the export source kind
+    description: Specify the export source kind (vm, vmsnapshot, pvc)
```
jcanocan, line 27: Same as above.

templates/disk-uploader/manifests/disk-uploader.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     to be used in multiple Kubernetes clusters.
   params:
   - name: EXPORT_SOURCE_KIND
-    description: The name of the export source kind
+    description: Specify the export source kind (vm, vmsnapshot, pvc)
```
jcanocan, line 27: Same as above.

release/tasks/disk-uploader/README.md
codingben: I'm not sure about this. `vm`, `vmsnapshot` or `pvc` is the expected value. It won't work if the user will set another value than [1].

[1] https://github.com/kubevirt/kubevirt-tekton-tasks/blob/main/modules/disk-uploader/pkg/vmexport/vmexport.go#L21-L25

jcanocan: Ah I see, so those are the expected values of the field. Maybe you can clarify it, at least confused me.

codingben:
@ksimon1 @jcanocan Can you please review it again?

templates/disk-uploader/readmes/README.md
```diff
@@ -18,6 +18,13 @@ When user runs [KubeVirt Tekton Tasks](https://github.com/kubevirt/kubevirt-tekt
 
 ### Usage
 
+Secret of the task run:
```
jcanocan, line 21:
```suggestion
To get the Secret of the task run:
```

release/tasks/disk-uploader/README.md
```diff
@@ -12,7 +12,7 @@ When user runs [KubeVirt Tekton Tasks](https://github.com/kubevirt/kubevirt-tekt
 
 ### Parameters
 
-- **EXPORT_SOURCE_KIND**: The name of the export source kind
+- **EXPORT_SOURCE_KIND**: Specify the export source kind (Expected values `vm`, `vmsnapshot`, or `pvc`)
 - **EXPORT_SOURCE_NAME**: The name of the export source
```
geetikakay, line 16: I believe we should mention push_timeout is optional in default value in readme

codingben: Ok, mentioned it.

PR #597, opened 2025-01-15
fix(CNV-54663): show status of used export sources

codingben:

Currently disk-uploader will always show
ready status, and there is no indication
of pending status that could mean of an
used export source (e.g. an used PVC that
is not available to be exported).

**Release note**:

```release-note
Show status of used export sources.
```

### DIFF
--- a/modules/disk-uploader/pkg/vmexport/vmexport.go
+++ b/modules/disk-uploader/pkg/vmexport/vmexport.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/kubevirt/kubevirt-tekton-tasks/modules/shared/pkg/log"
 	"github.com/kubevirt/kubevirt-tekton-tasks/modules/shared/pkg/ownerreference"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	"go.uber.org/zap"
 
 	kvcorev1 "kubevirt.io/api/core/v1"
 	v1beta1 "kubevirt.io/api/export/v1beta1"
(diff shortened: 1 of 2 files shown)


codingben:
@ksimon1 @geetikakay Can you please take a look?

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -57,8 +60,17 @@ func WaitUntilVirtualMachineExportReady(client kubecli.KubevirtClient, namespace
 			return false, err
 		}
 
-		if vmExport.Status != nil && vmExport.Status.Phase == v1beta1.Ready {
-			return true, nil
+		if vmExport.Status != nil {
+			log.Logger().Info("VirtualMachineExport object status", zap.String("status", string(vmExport.Status.Phase)))
```
0xFelix, line 64: Is `zap.String` required here? What does it help with?

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -57,8 +60,17 @@ func WaitUntilVirtualMachineExportReady(client kubecli.KubevirtClient, namespace
...
+
+			if vmExport.Status.Phase == v1beta1.Ready {
+				log.Logger().Info("VirtualMachineExport is in Ready state, and export source is not longer used")
+				return true, nil
+			}
+		}
+
+		if vmExport.Status.Phase == v1beta1.Pending {
```
0xFelix, line 72: `vmExport.Status.Phase` can be nil here?

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: We already using it here [1] to log the namespace and name, so I thought it would be useful here as well.

[1] https://github.com/kubevirt/kubevirt-tekton-tasks/blob/main/cmd/disk-uploader/main.go#L38

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: Yeah missed it.

modules/disk-uploader/pkg/vmexport/vmexport.go
0xFelix: Hm, I see...

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -57,8 +60,18 @@ func WaitUntilVirtualMachineExportReady(client kubecli.KubevirtClient, namespace
...
+			log.Logger().Info("VirtualMachineExport object status", zap.String("status", string(vmExport.Status.Phase)))
+
+			if vmExport.Status.Phase == v1beta1.Ready {
+				log.Logger().Info("VirtualMachineExport is in Ready state, and export source is not longer used")
+				return true, nil
+			}
+
+			if vmExport.Status.Phase == v1beta1.Pending {
+				log.Logger().Info("VirtualMachineExport is in Pending state, and export source is used")
+				return false, nil
```
0xFelix, line 73: nit: It's okay to drop this line.

0xFelix (commented):
/approve

modules/disk-uploader/pkg/vmexport/vmexport.go
codingben: I know, just wanted to keep consistency between if conditions. Let's leave it.

PR #614, opened 2025-01-29
feat: do not exit with error when tekton result folder does not exists in disk-uploader

ksimon1:
feat: do not exit with error when tekton result folder does not exists in disk-uploader

Disk uploader binary can be triggered via Tekton Tasks or via OCP UI in
disk uploader feature. When UI triggers the binary, binary is looking for
Tekton result folder, which does not exist and ends with error.
If the Tekton result folder is not found, the binary will just exit with
0.
**Release note**:
```
feat: do not exit with error when tekton result folder does not exists in disk-uploader
```

### DIFF
--- a/cmd/disk-uploader/main.go
+++ b/cmd/disk-uploader/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"errors"
+	"io/fs"
 	"os"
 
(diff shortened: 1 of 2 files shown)


codingben (approved):
/lgtm

PR #649, opened 2025-02-05
[release-v0.16] fix:insert tag in common templates URL in deploy resource script

ksimon1:
fix:insert tag in common templates URL in deploy resource script

change ssp api version

update version of fedora template used in e2e tests

**Release note**:
```release-note
NONE
```


ksimon1:
Failed test is not caused by this PR

PR #652, opened 2025-02-10
feat: increate poll interval from 15 seconds to 60 seconds

ksimon1:
feat: increate poll interval from 15 seconds to 60 seconds

the 15 seconds poll intervall is too short and if user tries to export bigger disk, it puts too much messages in the log. The value is increased to 60 seconds

```release-note
NONE
```


geetikakay:
lgtm

0xFelix (commented):
/approve

PR #666, opened 2025-02-26
fix: test-with-reports makefile target

ksimon1:
fix: test-with-reports makefile target

the script was checking if module contains go.mod file and if not it skips the module. The go.mod files were removed and only one is kept in the root of the project. This PR removes the check for go.mod file and fixes failing unit tests.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/modules/disk-uploader/pkg/vmexport/vmexport_test.go
+++ b/modules/disk-uploader/pkg/vmexport/vmexport_test.go
@@ -4,23 +4,22 @@ import (
 	"context"
 	"os"
 
+	"github.com/golang/mock/gomock"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"github.com/golang/mock/gomock"
-
+	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


modules/disk-uploader/pkg/vmexport/vmexport_suite_test.go
```diff
@@ -3,11 +3,14 @@ package vmexport_test
...
 	"testing"
 
+	"github.com/kubevirt/kubevirt-tekton-tasks/modules/shared/pkg/log"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"go.uber.org/zap"
 )
 
 func TestVMExport(t *testing.T) {
+	log.InitLogger(zap.InfoLevel)
```
0xFelix, line 13: Do we require the logger here?

ksimon1: Yes, otherwise the vmexport unit test panics

jcanocan: What do you think about including a comment explaining it?

0xFelix: Can you add it to the test that requires it?

0xFelix (approved):
/lgtm

PR #726, opened 2025-05-28
fix: update example windows customize configmap

ksimon1:
fix: update example windows customize configmap

Latest Windows 11 24h2 contains bug, which is causing the VM to boot into BSOD. This is happening because, when VM is syspreped and bitlocker is enabled, bitlocker modifies bootloader parameters, which then forces VM to boot into BSOD. This PR is updating the Win11 unattend xml and adding commands which fixes the parameters for bootloader.

**Release note**:
```
NONE
```

### DIFF
--- a/templates-pipelines/windows-customize/configmaps/windows-customize-configmaps.yaml
+++ b/templates-pipelines/windows-customize/configmaps/windows-customize-configmaps.yaml
@@ -245,9 +245,17 @@ data:
           </AutoLogon>
           <FirstLogonCommands>
             <SynchronousCommand wcm:action="add">
+              <CommandLine>bcdedit /set device partition=C:</CommandLine>
+              <Order>1</Order>
+            </SynchronousCommand>
+            <SynchronousCommand wcm:action="add">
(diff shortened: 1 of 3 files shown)


dominikholler:
The updated windows11-unattend fixed for me the issue on failed reboot on bitlocker enablement.

templates-pipelines/windows-customize/configmaps/windows-customize-configmaps.yaml
```diff
@@ -244,10 +244,18 @@ data:
             <LogonCount>1</LogonCount>
           </AutoLogon>
           <FirstLogonCommands>
+            <SynchronousCommand wcm:action="add">
+              <CommandLine>bcdedit /set device partition=C:</CommandLine>
+              <Order>1</Order>
+            </SynchronousCommand>
+            <SynchronousCommand wcm:action="add">
+              <CommandLine>bcdedit /set osdevice partition=C:</CommandLine>
+              <Order>2</Order>
```
geetikakay, line 253: optional but is there a need to add false and description for two new commands?

PR #728, opened 2025-06-06
feat: update unattend.xml for win11 example with official MS fix

ksimon1:
feat: update unattend.xml from win11 example with official MS fix

https://learn.microsoft.com/en-us/answers/questions/1843393/windows-11-24h2-26100-1150-sysprep-generalize-brea?source=docs 
This should be official MS fix for win11 bsod caused by bitlocker

**Release note**:
```release-note
feat: update unattend.xml from win11 example with official MS fix

```

### DIFF
--- a/release/pipelines/windows-customize/configmaps/windows-customize-configmaps.yaml
+++ b/release/pipelines/windows-customize/configmaps/windows-customize-configmaps.yaml
@@ -245,20 +245,36 @@ data:
           </AutoLogon>
           <FirstLogonCommands>
             <SynchronousCommand wcm:action="add">
-              <CommandLine>bcdedit /set device partition=C:</CommandLine>
+              <CommandLine>PowerShell -ExecutionPolicy Bypass -NoProfile F:\post-update.ps1</CommandLine>
               <Order>1</Order>
-            </SynchronousCommand>
-            <SynchronousCommand wcm:action="add">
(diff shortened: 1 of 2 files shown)


jcanocan:
Nice job, and thanks for your persistence with this!
It looks good to me, nothing to add regarding the xml modifications.
However, I would be a bit more descriptive with the NOTES. E.g.,
```
feat: Add fix for BSOD in Windows 11 24H2 when Bitlocker is enabled
```

0xFelix:
Not sure I like the wording of "official fix", since this needs to be fixed  within Windows afterall, but it's a viable workaround.

/approve

release/pipelines/windows-customize/configmaps/windows-customize-configmaps.yaml
```diff
@@ -245,20 +245,36 @@ data:
           </AutoLogon>
           <FirstLogonCommands>
             <SynchronousCommand wcm:action="add">
-              <CommandLine>bcdedit /set device partition=C:</CommandLine>
+              <CommandLine>PowerShell -ExecutionPolicy Bypass -NoProfile F:\post-update.ps1</CommandLine>
```
kostyanf14, line 248: In general, I preferred to enumerate all drives for the post-update script, instead of hardcoding F:. F: means that you exactly know that Windows has only one drive C:, the second drive D: is a Windows ISO, etc...

PR #744, opened 2025-08-28
[release-v0.15] fix: hardcode ssp operator to version v0.18.3

ksimon1:
fix: hardcode ssp operator to version v0.18.3

Newest SSP operator contains operands, which do
not work on older OCP versions

**Release note**:
```release-note
NONE
```


0xFelix (commented):
Same as in other PR, would it make sense to hardcode Kubevirt and CDI too?

/lgtm
/hold

0xFelix (approved):
/lgtm

PR #847, opened 2026-05-06
fix: switch VS Code to system-level installer in Windows customize pipeline

ksimon1:
Use the system installer (win32-x64) instead of the user installer (win32-x64-user) and add AppxPackage cleanup after installation.
Our CI is failing due to the error during syspreping in customize pipeline. This PR will fix it

**Release note**:
```release-note
fix: switch VS Code to system-level installer in Windows customize pipeline
```


0xFelix:
What does this fix? Can you explain? Also, it's an example anyway.

ksimon1:
> Ou CI is failing due to the error during syspreping in customize pipeline. This PR will fix it

Eventhough it is an example, we are using it in CI in pipeline tests :)
I added note, that this PR will fix our CI

0xFelix (approved):
/lgtm

Issue #846, opened 2026-05-05
RFE: Add download and push progress reporting to disk-uploader


galkremer1:
## Summary

The `disk-uploader` task currently provides no progress feedback during the two most time-consuming phases of the upload pipeline: **disk image download** and **container image push to registry**. Users launching the "Upload to registry" workflow from the OpenShift Console kubevirt plugin have no way to estimate how long the operation will take or distinguish a slow upload from a hung one.

## Problem

The pipeline has four distinct stages:

1. Create `VirtualMachineExport` and wait for it to become `Ready`
2. **Download disk image** via `nbdkit` + `qemu-img convert` (can be many GB — no progress)
3. Build OCI layer from disk image
4. **Push container image to registry** via `crane.Push` (can be many GB — no progress)

During stages 2 and 4 the pod stays in `Running` phase with no observable state change, making it impossible for external tooling or UIs to report meaningful progress.

## Proposed Solution

Write structured progress updates to a Kubernetes **ConfigMap** (named after the pod, in the same namespace) so that any watcher — including the kubevirt-plugin UI — can display real-time progress without parsing unstructured logs.

### ConfigMap shape (example)
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: <pod-name>   # e.g. kubevirt-disk-uploader-abc123
  namespace: <ns>
data:
  phase: "downloading"   # waiting | downloading | building | pushing | done
  percent: "42"
  bytesComplete: "1073741824"
  bytesTotal: "2684354560"
```

### Implementation sketch

**Download phase (`disk.go`):**
- Add `-p` flag to `qemu-img convert` which emits progress to stderr
- Parse the stderr output in a goroutine and patch the ConfigMap

**Push phase (`image.go`):**
- `go-containerregistry` already supports progress via `crane.WithProgress(chan v1.Update)` where `v1.Update{Complete, Total int64}` carries byte counts — this is a one-liner
- A goroutine reads the channel and patches the ConfigMap

**RBAC:**
- Add `configmaps: [create, update, delete]` to the uploader's Role

**Cleanup:**
- Delete the ConfigMap when the pod exits (success or failure)

### UI side (kubevirt-plugin)
Once available, kubevirt-plugin can watch the ConfigMap by pod name and render a real progress bar instead of the current static stepper step.

## Motivation

During testing of the Upload to Registry modal ([CNV-71939](https://redhat.atlassian.net/browse/CNV-71939)), we observed:
[...]

PR #855, opened 2026-05-13
feat(disk-uploader): add download and upload progress logging

codingben:

Add structured JSON log output for tracking progress when
downloading the VM disk from the source and uploading it
to the container registry.

The following fields are logged every 5% threshold:

- percentage (download and upload)
- bytes_uploaded / total_bytes
- speed_mbps
- time_remaining

This allows the UI to stream pod logs and parse the JSON
entries for real-time progress display without requiring
additional RBAC permissions.

This PR will address #846 issue.

**Release note**:

```release-note
Add download and upload progress logging.
```


modules/disk-uploader/pkg/image/image.go
```diff
@@ -64,11 +65,46 @@ func Push(image v1.Image, imageDestination string, pushTimeout int) error {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*time.Duration(pushTimeout))
 	defer cancel()
 
+	ref, err := name.ParseReference(imageDestination)
+	if err != nil {
+		log.Fatalf("Error parsing image reference: %v", err)
```
0xFelix: Bot has a point, also this changes looks to be unrelated to the progress indicator?

ksimon1:
e2e test are failing with:
```
  panic: close of closed channel
  goroutine 1 [running]:
  github.com/kubevirt/kubevirt-tekton-tasks/modules/disk-uploader/pkg/image.Push({0x254c350, 0xc0003fe100}, {0x7ffda2a50c55, 0x1f}, 0x0?)
  	/src/modules/disk-uploader/pkg/image/image.go:101 +0x374
  main.run({{0x7ffda2a50bef, 0x2}, {0xc00005800e, 0xf}, {0x7ffda2a50c07, 0xd}, {0x7ffda2a50c22, 0x1f}, {0x7ffda2a50c55, 0x1f}, ...}, ...)
  	/src/cmd/disk-uploader/main.go:104 +0x9e5
  main.main()
  	/src/cmd/disk-uploader/main.go:143 +0x40d
  << Captured StdOut/StdErr Output
  [FAILED] Unexpected error:
      <*errors.errorString | 0xc00128d7b0>: 
      "e2e-tests-taskrun-disk-uploader-k65h6" failed
      {
          s: "\"e2e-tests-taskrun-disk-uploader-k65h6\" failed",
      }
```

modules/disk-uploader/pkg/image/image.go
codingben: Yes right. `name.ParseReference` is required to get the ref.

codingben:
> panic: close of closed channel

It shouldn't be panic now. I thought it's a bug in the library itself.

codingben:
This is a very good note from the bot above:

> The upload progress logger in Push logs on every progress update, which could be very noisy for large images; consider adding simple throttling (e.g., only logging on whole-percentage changes or at a fixed interval).

It definitely spams the log with the output of percentage during download and upload, @0xFelix @ksimon1 How will you suggest to adjust it or leave it as it is?

codingben:
> This is a very good note from the bot above:
> 
> > The upload progress logger in Push logs on every progress update, which could be very noisy for large images; consider adding simple throttling (e.g., only logging on whole-percentage changes or at a fixed interval).
> 
> It definitely spams the log with the output of percentage during download and upload, @0xFelix @ksimon1 How will you suggest to adjust it or leave it as it is?

The issue that it's not easily achievable by qemu during download, and upload is also would look weird if we'll have less logs (let's say output each 10%).

ksimon1:
@codingben Will something like this https://github.com/kubevirt/kubevirt-tekton-tasks/commit/a485fecf372cb3c752542ba8d3831766d7dfdd82 do the trick?

codingben:
> @codingben Will something like this [a485fec](https://github.com/kubevirt/kubevirt-tekton-tasks/commit/a485fecf372cb3c752542ba8d3831766d7dfdd82) do the trick?

I had similar implementation and don't like using Regex here. I think Regex is more-fit to one time checking than each frame.

Let's simplify it, do we want to spam it with logs every other 1% percentage update, or limit it to log every 10% percentage update?

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +61,85 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
 	return image, nil
 }
 
-func Push(image v1.Image, imageDestination string, pushTimeout int) error {
+func Push(image v1.Image, imageDestination string, pushTimeout int, onProgress func(float64)) error {
```
ksimon1: please divide this function into smaller fns

cmd/disk-uploader/main.go
```diff
@@ -37,6 +49,18 @@ func run(opts parse.CLIOptions, k8sClient kubernetes.Interface, virtClient kubec
 	imageDestination := opts.GetImageDestination()
 	imagePushTimeout := opts.GetPushTimeout()
 
+	podName := os.Getenv("POD_NAME")
```
ksimon1: please add these values in parse.CLIOptions struct

cmd/disk-uploader/main.go
```diff
@@ -29,6 +33,14 @@ const (
 	kvExportTokenHeader string = "x-kubevirt-export-token"
 )
 
+func patchPodAnnotation(k8sClient kubernetes.Interface, namespace, podName, key, value string) {
```
ksimon1: how often is this function called?

codingben: Every increase of 10% in download/upload progress.

modules/disk-uploader/pkg/image/image.go
codingben: Yes can be smaller. Done

cmd/disk-uploader/main.go
codingben: Done

cmd/disk-uploader/main.go
galkremer1: @codingben so we are only going to be able to track the progress in 10% increments?
Is there a way for us to estimate the remaining time of the upload?

codingben: @galkremer1 Yes, from UI perspective you can nicely increase the progress bar from X to Y value. There is no way of calculating the time of the whole process.

galkremer1: Thanks. Is there a way for you to expose the current upload speed using the annotation?

galkremer1: It would be best if the estimated remaining time can be calculated on your end and we will also get an indication of the upload speed for best UX.

codingben: @galkremer1 I'm working on this.

codingben: @galkremer1 I've added both annotations to show the time left in seconds and network speed in mb/s:
```
disk-uploader.kubevirt.io/upload-speed: 14.24 MB/s
disk-uploader.kubevirt.io/upload-time-remaining: 5s
```
All the annotations (and logs) will be updated in each increase of 5% to reduce k8s API calls by default, but you can adjust it via setting `PROGRESS_THRESHOLD` env var.

codingben:
@galkremer1 I've added the new annotations of upload speed and time left in seconds. @ksimon1 Can you please review this?

modules/disk-uploader/pkg/constants/constants.go
```diff
@@ -1,12 +1,28 @@
...
 	InvalidCLIInputExitCode   = 2
 	DiskUploaderErrorExitCode = 3
 	WriteResultsExitCode      = 4
 )
 
+func ProgressThreshold() float64 {
+	const defaultProgressThreshold = 5.0
+
+	if v := os.Getenv("PROGRESS_THRESHOLD"); v != "" {
+		if t, err := strconv.ParseFloat(v, 64); err == nil && t > 0 {
```
ksimon1: Accepts any value > 0, including values > 100 which don't make sense for a percentage threshold.

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +68,107 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
...
+		}
+		totalSize += size
+	}
+
+	log.Logger().Info("Image size calculated", zap.Int64("total_bytes", totalSize), zap.Int("layer_count", len(layers)))
+	return totalSize, nil
+}
+
+func registryAuth() *authn.Basic {
+	return &authn.Basic{
```
ksimon1: It would make sense to check if the values exists and fail if they are empty

cmd/disk-uploader/main.go
galkremer1: Thanks, that's great.

galkremer1 (approved):
/lgtm

modules/disk-uploader/pkg/constants/constants.go
codingben: Done

modules/disk-uploader/pkg/image/image.go
codingben: Done

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +68,122 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
...
+			if update.Error != nil {
+				log.Logger().Error("Upload error", zap.Error(update.Error))
+				continue
+			}
+
+			if update.Complete > 0 && totalSize > 0 {
+				percentage := float64(update.Complete) / float64(totalSize) * 100
+				if percentage-lastLogged >= threshold || percentage >= 100.0 {
+					elapsed := time.Since(startTime).Seconds()
+					speedBps := float64(update.Complete) / elapsed
```
ksimon1: Is there a chance, that elapsed might be a zero?

cmd/disk-uploader/main.go
```diff
@@ -29,13 +34,36 @@ const (
...
+	podName := opts.GetPodName()
+	podNamespace := opts.GetPodNamespace()
+
+	onDownloadProgress := func(pct float64) {
+		patchPodAnnotation(k8sClient, podNamespace, podName,
+			"disk-uploader.kubevirt.io/download-progress", fmt.Sprintf("%.1f", pct))
+	}
+	onUploadProgress := func(p image.UploadProgress) {
+		patchPodAnnotation(k8sClient, podNamespace, podName,
+			"disk-uploader.kubevirt.io/upload-progress", fmt.Sprintf("%.1f", p.Percentage))
```
ksimon1: please use constants for the annotations

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +68,122 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
...
+		return fmt.Errorf("error parsing image reference: %v", err)
+	}
+
+	totalSize, err := calculateImageSize(image)
+	if err != nil {
+		return err
 	}
-	err := crane.Push(image, imageDestination, crane.WithAuth(auth), crane.WithContext(ctx))
+
+	auth, err := registryAuth()
```
ksimon1: This function is worth to be called at the beginning of disk uploader, so we fail early

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -4,9 +4,51 @@ import (
...
+	"go.uber.org/zap"
 )
 
-func DownloadDiskImageFromURL(rawDiskUrl, headerKey, headerValue, certificatePath, diskPath string) error {
+// progressThrottler implements io.Writer to throttle progress logging
+type progressThrottler struct {
+	re           *regexp.Regexp
+	lastLogged   float64
+	threshold    float64
+	outputToUser bool
```
ksimon1: is this variable used anywhere?

modules/disk-uploader/pkg/constants/constants.go
```diff
@@ -1,12 +1,28 @@
...
 	InvalidCLIInputExitCode   = 2
 	DiskUploaderErrorExitCode = 3
 	WriteResultsExitCode      = 4
 )
 
+func ProgressThreshold() float64 {
+	const defaultProgressThreshold = 5.0
+
+	if v := os.Getenv("PROGRESS_THRESHOLD"); v != "" {
+		if t, err := strconv.ParseFloat(v, 64); err == nil && t > 0 && t <= 100 {
```
ksimon1: I think we should have here stricter check. I can set e.g. 0.0000000000001 and then log and api will explode :)

codingben: Done

modules/disk-uploader/pkg/disk/disk.go
codingben: Nope

modules/disk-uploader/pkg/image/image.go
codingben: Done

cmd/disk-uploader/main.go
codingben: Done

modules/disk-uploader/pkg/image/image.go
codingben: Yes, resolved

codingben:
@ksimon1 Thanks for your review comments. Can we merge this? I think we're good here.

galkremer1 (commented):
/lgtm

codingben:
> @galkremer1: changing LGTM is restricted to collaborators
> 
> Details

@dhiller @ksimon1 Why Gal can't add lgtm label? He's part of KV org and should be able to add such label in all repos under this org. No?

galkremer1:
> > @galkremer1: changing LGTM is restricted to collaborators
> > Details
> 
> @dhiller @ksimon1 Why Gal can't add lgtm label? He's part of KV org and should be able to add such label in all repos under this org. No?

I am a part of the Kubevirt-ui org, not Kubevirt.

cmd/disk-uploader/main.go
```diff
@@ -27,15 +32,48 @@ const (
...
+	if err != nil {
+		return "", err
+	}
+
+	onDownloadProgress := func(pct float64) {
+		patchPodAnnotation(k8sClient, podNamespace, podName,
+			annotationDownloadProgress, fmt.Sprintf("%.1f", pct))
+	}
+	onUploadProgress := func(p image.UploadProgress) {
+		patchPodAnnotation(k8sClient, podNamespace, podName,
```
ksimon1: Please do just a single patch call, instead of 3 different calls.

cmd/disk-uploader/main.go
```diff
@@ -27,15 +32,48 @@ const (
...
+
+	annotationDownloadProgress = "disk-uploader.kubevirt.io/download-progress"
+	annotationUploadProgress   = "disk-uploader.kubevirt.io/upload-progress"
+	annotationUploadSpeed      = "disk-uploader.kubevirt.io/upload-speed"
+	annotationUploadTimeRemain = "disk-uploader.kubevirt.io/upload-time-remaining"
 )
 
+func patchPodAnnotation(k8sClient kubernetes.Interface, namespace, podName, key, value string) {
+	patch := fmt.Sprintf(`{"metadata":{"annotations":{%q:%q}}}`, key, value)
+	_, err := k8sClient.CoreV1().Pods(namespace).Patch(context.Background(), podName, types.MergePatchType, []byte(patch), metav1.PatchOptions{})
```
ksimon1: please use e.g. 
```
ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
defer cancel()
_, err := k8sClient.CoreV1().Pods(namespace).Patch(ctx, podName, types.MergePatchType, []byte(patch), metav1.PatchOptions{})
```

cmd/disk-uploader/main.go
codingben: Good catch. Let's simplify this.

cmd/disk-uploader/main.go
codingben: Done

codingben:
@ksimon1 Thanks for your comments. I've resolved the last ones.

ksimon1:
One last comment, can you enhance the disk-uploader e2e test and add check if the annotations are set on the pod and are showing something?

codingben:
> One last comment, can you enhance the disk-uploader e2e test and add check if the annotations are set on the pod and are showing something?

Done. I think it's more useful by UI, than in a Tekton, since only UI will be using it.

akrejcir:
I have a design question: Have you considered adding an HTTP server into the `disk-uploader` pod, that would return the progress information? The UI could keep a connection to it open, or poll it regularly. 
I'm not very familiar with the code, so I don't know how realistic it would be, or if it is a good design choice.

codingben:
> I have a design question: Have you considered adding an HTTP server into the `disk-uploader` pod, that would return the progress information? The UI could keep a connection to it open, or poll it regularly. I'm not very familiar with the code, so I don't know how realistic it would be, or if it is a good design choice.

No. The implementation of `disk-uploader` was designed to work in a Tekton pipeline (e.g. Windows one) and the pod is short-lived just to perform upload actions once deployed. Not every cluster is coming with Tekton pre-installed, and therefore there is just a UI button to deploy this pod directly by the user (it can be also deployed by kubecli of course.).

akrejcir:
I think you maybe misunderstood my question.
I meant that the short lived Pod would have an open port with an HTTP service listening on it. So then the UI could connect directly to the Pod's IP. But there may be issues with network policies, so I'm unsure if the connection would be allowed.

ksimon1:
/approve
/hold
Giving @0xFelix time to review this PR
@akrejcir I would like to avoid the http server, because then we would have to solve the security of the server

codingben:
> I think you maybe misunderstood my question. I meant that the short lived Pod would have an open port with an HTTP service listening on it. So then the UI could connect directly to the Pod's IP. But there may be issues with network policies, so I'm unsure if the connection would be allowed.

The UI can't connect directly to the Pod's IP because it's just a based-JS web browser code. It would need be exposed outside the cluster. The UI can watch the pod's annotations via k8s API and get updates without opening additional connections (`/watch` already opens one streaming connection).

akrejcir:
> @akrejcir I would like to avoid the http server, because then we would have to solve the security of the server

Good point.

> The UI can't connect directly to the Pod's IP because it's just a based-JS web browser code. It would need be exposed outside the cluster. The UI can watch the pod's annotations via k8s API and get updates without opening additional connections (/watch already opens one streaming connection).

Right, I did not consider that. You can ignore my comment.

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -4,9 +4,49 @@ import (
...
+// progressThrottler implements io.Writer to throttle progress logging
+type progressThrottler struct {
+	re         *regexp.Regexp
+	lastLogged float64
+	threshold  float64
+	onProgress func(float64)
+}
+
+func (pt *progressThrottler) Write(p []byte) (n int, err error) {
+	line := string(p)
```
akrejcir: The `Writer` is probably not guaranteed to receive a full line in the parameter `p`. It could be a partial line, or multiple lines.

Can you add code that would handle these cases? It could accumulate bytes in `p` until it receives a full line. There may already exist a wrapper in the library that does that.

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -4,9 +4,49 @@ import (
...
+	for _, match := range matches {
+		if len(match) > 1 {
+			if pct, err := strconv.ParseFloat(match[1], 64); err == nil {
+				if pct-pt.lastLogged >= pt.threshold || pct >= 100.0 {
+					log.Logger().Info("Downloading disk image", zap.Float64("percentage", pct))
+
+					pt.lastLogged = pct
+
+					if pt.onProgress != nil {
+						pt.onProgress(pct)
```
akrejcir: `onProgress()` calls the API and can potentially take a long time. If the logs are faster than the API calls, then they accumulate waiting for the writer to consume them. For the progress reporting, only the latest percentage is relevant.

Can you decouple the progress reporting from the log parsing?
The parser could store the latest value into a variable and then wake up a goroutine that would call `onProgress()`.

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -15,9 +55,14 @@ func DownloadDiskImageFromURL(rawDiskUrl, headerKey, headerValue, certificatePat
 		fmt.Sprintf("header=%s: %s", headerKey, headerValue),
 		fmt.Sprintf("cainfo=%s", certificatePath),
 		"--run",
-		fmt.Sprintf("qemu-img convert \"$uri\" -O qcow2 %s", diskPath),
+		fmt.Sprintf("qemu-img convert \"$uri\" -p -O qcow2 %s", diskPath),
 	)
-	cmd.Stdout = os.Stdout
+	cmd.Stdout = &progressThrottler{
```
akrejcir: Previously, the output was written to `os.Stdout`, but now it is swallowed by the `progressThrottler`, and not written anywhere.

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +69,117 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
...
 	}
-	err := crane.Push(image, imageDestination, crane.WithAuth(auth), crane.WithContext(ctx))
+
+	totalSize, err := calculateImageSize(image)
 	if err != nil {
-		log.Fatalf("Error pushing image: %v", err)
 		return err
 	}
+
+	progressChan := make(chan v1.Update, 100)
```
akrejcir, line 83: The allocated size 100 is unusually large. Can you add a comment why did you choose this number?

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +69,117 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
...
+	done := trackUploadProgress(progressChan, totalSize, onProgress)
```
akrejcir: It is more idiomatic to separate goroutine creation from functionality. Can you rewrite this as:
```go
go func() {
    defer close(done)
    trackUploadProgress(progressChan, totalSize, onProgress)
}()
```

And remove the goroutine creation from the `trackUploadProgress`?

modules/disk-uploader/pkg/image/image.go
```diff
@@ -60,18 +69,117 @@ func Build(diskPath string, config v1.Config) (v1.Image, error) {
...
+						zap.Float64("percentage", percentage),
+						zap.Int64("bytes_uploaded", update.Complete),
+						zap.Int64("total_bytes", totalSize),
+						zap.Float64("speed_mbps", speedMBps),
+						zap.String("time_remaining", timeRemaining.Truncate(time.Second).String()))
+
+					lastLogged = percentage
+
+					if onProgress != nil {
+						onProgress(UploadProgress{
```
akrejcir: Calling `onProgress` can potentially take a longer time, and events in `progressChan` may accumulate in the meantime. If there are multiple events accumulated in the channel, then only the last event is relevant for `onProgress()` call.

Can you change this so that received events only update a variable containing the latest state, and trigger a separate goroutine that reads this variable and calls `onProgress()` when the variable changed.

modules/disk-uploader/pkg/disk/disk.go
codingben: Good catch. Done.

0xFelix:
There's a lot of added complexity in this PR. Are we sure that updating the logs of the Pod is not sufficient? From what I understood this is a feature we are going to phase out anyway?

codingben:
> There's a lot of added complexity in this PR.

There is definitely more complexity in parsing the download progress. To parse the upload progres is simpler via `progressChan` below:
```
err = remote.Write(ref, image,
	remote.WithAuth(auth),
	remote.WithContext(ctx),
	remote.WithProgress(progressChan),
)
```
Both will update the annotations with values so UI can watch them.

> Are we sure that updating the logs of the Pod is not sufficient? From what I understood this is a feature we are going to phase out anyway?

This is not useful by UI team and will not resolve this issue. They want to have key/value where they can take the calculated values and display them to the user.

modules/disk-uploader/pkg/disk/disk.go
codingben: Where it should be written? The output is still displayed in the logs.

modules/disk-uploader/pkg/disk/disk.go
codingben: Done. Can you please review it again?

modules/disk-uploader/pkg/image/image.go
codingben: I've added it.

modules/disk-uploader/pkg/image/image.go
codingben: Yes.

modules/disk-uploader/pkg/image/image.go
codingben: Makes sense. Can you review it again?

0xFelix:
> This is not useful by UI team and will not resolve this issue. They want to have key/value where they can take the calculated values and display them to the user.

I also want to have a lot of things. Just because someone yells doesn't mean you have to do it.

Adding `patch` permissions for all Pods in a give namespace to a simple disk uploader doesn't sound like the best idea to me.

A simple change would be to emit the progress as logs on stdout of the Pod and to point users towards these logs in the UI. Why is there a need for such a complex solution?

There's a good chance we will have a better solution in the future, which will be backed by a proper CRD and controller (`VirtualMachineExport`). The current solution mixes the concerns of view and controller, as it is implemented mainly in the UI. That's why I'm asking: Are we not going to phase this out anyway?

codingben:
> > This is not useful by UI team and will not resolve this issue. They want to have key/value where they can take the calculated values and display them to the user.
> 
> I also want to have a lot of things. Just because someone yells doesn't mean you have to do it.
> 
> Adding `patch` permissions for all Pods in a give namespace to a simple disk uploader doesn't sound like the best idea to me.
> 
> A simple change would be to emit the progress as logs on stdout of the Pod and to point users towards these logs in the UI. Why is there a need for such a complex solution?
> 
> There's a good chance we will have a better solution in the future, which will be backed by a proper CRD and controller (`VirtualMachineExport`). The current solution mixes the concerns of view and controller, as it is implemented mainly in the UI. That's why I'm asking: Are we not going to phase this out anyway?

Thank you. Let's just keep the logs. UI definitely can get the JSON-based logs and parse the relevant values they needed as described in the PR. I've removed the complexity of updating pod's annotations.

For example:
```
{"level":"info","ts":1779710497.2893703,"caller":"disk/disk.go:34","msg":"Downloading disk image","percentage":95.25}
{"level":"info","ts":1779710534.0653534,"caller":"image/image.go:151","msg":"Uploading image","percentage":15.08,"bytes_uploaded":90177887,"total_bytes":597932928,"speed_mbps":11.96,"time_remaining":"40s"}
```

@galkremer1 Please see the latest changes, and use your k8s client to read the pod's logs to get the required values so you can display them to the user as a progress bar.

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -15,9 +83,9 @@ func DownloadDiskImageFromURL(rawDiskUrl, headerKey, headerValue, certificatePat
 		fmt.Sprintf("header=%s: %s", headerKey, headerValue),
 		fmt.Sprintf("cainfo=%s", certificatePath),
 		"--run",
-		fmt.Sprintf("qemu-img convert \"$uri\" -O qcow2 %s", diskPath),
+		fmt.Sprintf("qemu-img convert \"$uri\" -p -O qcow2 %s", diskPath),
 	)
-	cmd.Stdout = os.Stdout
+	cmd.Stdout = pt
```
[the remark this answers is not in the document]
codingben: I think this is AI mistake, both me and @ksimon1 saw it. It's only works when setting `cmd.Stdout`.

codingben:
@akrejcir @0xFelix Thanks for your review comments, PR's complexity was reduced to keep just logging. Can you please review it again?

codingben:
I don't see that the failure is related to disk-uploader changes.

/retest-required

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -15,9 +83,9 @@ func DownloadDiskImageFromURL(rawDiskUrl, headerKey, headerValue, certificatePat
...
 	cmd.Stderr = os.Stderr
```
0xFelix: qemu-img `-p` writes its progress indicator to stderr, not stdout. With `cmd.Stdout = pt` the throttler never receives any progress data - download progress will silently not be logged.

Fix: connect stderr (or a tee of stderr) to the throttler, e.g.:
```go
cmd.Stdout = os.Stdout
cmd.Stderr = pt
```
or if you still want stderr visible:
```go
cmd.Stderr = io.MultiWriter(os.Stderr, pt)
```

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -1,12 +1,80 @@
...
+	pt.wg.Add(1)
+	go pt.scanLines(pr)
+
+	return pt
+}
+
+func (pt *progressThrottler) scanLines(r io.Reader) {
+	defer pt.wg.Done()
+
+	scanner := bufio.NewScanner(r)
```
0xFelix, line 46: Even after fixing the stdout/stderr wiring above, `bufio.NewScanner` splits on `\n` by default, but qemu-img's `-p` output uses `\r` (carriage return) to overwrite the progress line in place. The scanner will accumulate the entire output as one "line" until EOF, then log all progress thresholds at once after the download already finished.

Fix: provide a custom split function, e.g.:
```go
scanner.Split(func(data []byte, atEOF bool) (advance int, token []byte, err error) {
    if i := bytes.IndexByte(data, '\r'); i >= 0 {
        return i + 1, data[:i], nil
    }
    if atEOF && len(data) > 0 {
        return len(data), data, nil
    }
    return 0, nil, nil
})
```

modules/disk-uploader/pkg/constants/constants.go
```diff
@@ -1,12 +1,26 @@
...
 	InvalidCLIInputExitCode   = 2
 	DiskUploaderErrorExitCode = 3
 	WriteResultsExitCode      = 4
-)
 
-const (
 	DigestResultName = "digest"
 )
+
+func ProgressThreshold() float64 {
```
0xFelix: Not really a constant, but is there a better place for it?

0xFelix (commented):
Added some comments by AI, also I think @sourcery-bot's comments are valid. Did you test this?

modules/disk-uploader/pkg/disk/disk.go
codingben: From testing it was working only with setting `cmd.Stdout` and didn't work with `cmd.Stderr`, both me and @ksimon1 saw it.

I've also replied about this [here](https://github.com/kubevirt/kubevirt-tekton-tasks/pull/855#discussion_r3398468914) to the AI bot.

modules/disk-uploader/pkg/disk/disk.go
codingben: I'll check this.

modules/disk-uploader/pkg/constants/constants.go
codingben: Will move it.

modules/disk-uploader/pkg/disk/disk.go
codingben: Okay I've applied your function.

codingben:
I've just verified the usage of latest changes in a development environemnt in the OpenShift cluster. Everything worked as expected (logs correctly printed in a JSON format) and the exported VM was pushed to the container registry.

@0xFelix Can you please take another look?

codingben:
> @codingben: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> ci/prow/e2e-tests	[3d8e7c2](https://github.com/kubevirt/kubevirt-tekton-tasks/commit/3d8e7c2865f7bd8ff3734688340435ba40b9610a)	[link](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/kubevirt_kubevirt-tekton-tasks/855/pull-ci-kubevirt-kubevirt-tekton-tasks-main-e2e-tests/2067627130048483328)	true	`/test e2e-tests`
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=kubevirt&repo=kubevirt-tekton-tasks&pr=855). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:codingben).
> 
> Details

This failure is caused by another functional test that is not related to this PR. Either something is broken in the CI again, or the functional test itself.

codingben:
Seems like now the job [e2e-tests](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/kubevirt_kubevirt-tekton-tasks/855/pull-ci-kubevirt-kubevirt-tekton-tasks-main-e2e-tests/2068715312752103424) passed after I've rebased this PR. @0xFelix @ksimon1 Can you please review if all good?

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -1,12 +1,90 @@
...
 
+type progressThrottler struct {
+	pw         *io.PipeWriter
+	re         *regexp.Regexp
+	lastLogged float64
+	threshold  float64
+	wg         sync.WaitGroup
+}
+
+func newProgressThrottler(re *regexp.Regexp, lastLogged, threshold float64) *progressThrottler {
```
akrejcir: The `lastLogged` parameter seems not needed. Can you use some default negative value?

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -1,12 +1,90 @@
...
+
+	return pt
+}
+
+func (pt *progressThrottler) scanLines(r io.Reader) {
+	defer pt.wg.Done()
+
+	scanner := bufio.NewScanner(r)
+	scanner.Split(func(data []byte, atEOF bool) (advance int, token []byte, err error) {
+		if i := bytes.IndexByte(data, '\r'); i >= 0 {
```
akrejcir, line 49: Can you add a comment into the code, why `\r` was chosen as a split?

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -1,12 +1,90 @@
...
+		}
+		if atEOF && len(data) > 0 {
+			return len(data), data, nil
+		}
+		return 0, nil, nil
+	})
+	for scanner.Scan() {
+		line := scanner.Text()
+		matches := pt.re.FindAllStringSubmatch(line, -1)
+		for _, match := range matches {
```
akrejcir: Does the logs contain multiple regex matches on a single token returned by the scanner?

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -1,12 +1,90 @@
...
+}
+
+func (pt *progressThrottler) Close() {
+	pt.pw.Close()
+	pt.wg.Wait()
+}
+
 func DownloadDiskImageFromURL(rawDiskUrl, headerKey, headerValue, certificatePath, diskPath string) error {
+	pt := newProgressThrottler(
+		regexp.MustCompile(`\((\d+(?:\.\d+)?)/100%\)`),
```
akrejcir: Can you compile the regex only once, and then reuse it?
Please also add a comment showing how does the log line from qemu looks like, so we can see what the regex is matching.

modules/disk-uploader/pkg/disk/disk.go
```diff
@@ -15,9 +93,9 @@ func DownloadDiskImageFromURL(rawDiskUrl, headerKey, headerValue, certificatePat
 		fmt.Sprintf("header=%s: %s", headerKey, headerValue),
 		fmt.Sprintf("cainfo=%s", certificatePath),
 		"--run",
-		fmt.Sprintf("qemu-img convert \"$uri\" -O qcow2 %s", diskPath),
+		fmt.Sprintf("qemu-img convert \"$uri\" -p -O qcow2 %s", diskPath),
```
akrejcir: Nit: Can you use long parameter names, instead on `-p` ?

akrejcir (commented):
Functionality looks good. I only have minor comments about code style.

modules/disk-uploader/pkg/disk/disk.go
codingben: Done.

modules/disk-uploader/pkg/disk/disk.go
codingben: I've added it.

modules/disk-uploader/pkg/disk/disk.go
codingben: No. `qemu-img` uses `\r` to overwrite progress in-place (e.g. `(0.00/100%)\r(1.50/100%)\r...`). The scanner splits on `\r` so each token contains only one percentage match. Simplified to `FindStringSubmatch`.

modules/disk-uploader/pkg/disk/disk.go
codingben: Done. Looks much better now.

modules/disk-uploader/pkg/disk/disk.go
codingben: Yep will be better.

codingben:
@akrejcir I've addressed your review comments and verified the changes in the cluster. Can you please review?

0xFelix (commented):
/approve

codingben:
@0xFelix @akrejcir Thank you! appreciate your help here with this complicated PR.

PR #936, opened 2026-07-23
fix: reduce flakiness in too-low-timeout execute/cleanup VM e2e tests

ksimon1:
fix: reduce flakiness in too-low-timeout execute/cleanup VM e2e tests

Two too-low-timeout entries were flaking because they inferred VM/SSH readiness from a guessed elapsed-time margin instead of an actual condition:
- "execute and stops vm with too low timeout" now polls for the VMI to reach Running (via Eventually) before starting its 27s timeout window, instead of assuming a fixed pre-wait was enough.
- "starts and execute and stops vm with too low timeout" now uses a 2s timeout, which is structurally too short for a fresh VMI to ever reach Running, instead of a value that was merely "usually" too low.

Also drop the dead "echo fail; exit 5" tail from the shared sleep script, since both tests rely on the process being killed mid-sleep and never reach it.

Assisted-by: Claude Sonnet 5

**Release note**:
```release-note
NONE
```

### DIFF
--- a/test/execute_and_cleanup_vm_test.go
+++ b/test/execute_and_cleanup_vm_test.go
@@ -23,16 +23,19 @@ const (
 	helloWorldScript = `#!/bin/bash
 echo hello world
 `
-	failScript = `#!/bin/bash
-echo fail
-exit 25
-`
+
 	sleepScript = `#!/bin/bash
 sleep 30
-echo fail
-exit 5
 `
 )
+
+func waitForVMRunning(f *framework.Framework, namespace, name string) {
+	Eventually(func(g Gomega) {
+		vmi, err := f.KubevirtClient.VirtualMachineInstance(namespace).Get(context.Background(), name, metav1.GetOptions{})
+		g.Expect(err).ShouldNot(HaveOccurred())
+		g.Expect(vmi.Status.Phase).To(Equal(kubevirtv1.Running))
+	}, Timeouts.WaitForVMStart.Duration, PollInterval).Should(Succeed())
+}
 
 var _ = Describe("Execute in VM / Cleanup VM", func() {
 	f := framework.NewFramework()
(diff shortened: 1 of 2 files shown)


test/execute_and_cleanup_vm_test.go
```diff
@@ -23,17 +23,25 @@ const (
...
 
+// waitForVMRunning polls until the VMI reaches the Running phase, instead of sleeping a
+// guessed duration. Fedora cloud images boot (and become SSH-reachable) within a few
+// seconds of the VMI going Running.
+func waitForVMRunning(f *framework.Framework, namespace, name string) {
+	Eventually(func() (kubevirtv1.VirtualMachineInstancePhase, error) {
+		vmi, err := f.KubevirtClient.VirtualMachineInstance(namespace).Get(context.Background(), name, metav1.GetOptions{})
+		if err != nil {
+			return "", err
+		}
```
0xFelix: Pass `g Gomega` and make `g.Expect` assertions?

test/execute_and_cleanup_vm_test.go
```diff
@@ -449,10 +462,15 @@ var _ = Describe("Execute in VM / Cleanup VM", func() {
...
-				Script:  sleepScript,
-				Timeout: constants.Timeouts.TenSeconds,
+				VM:     testobjects.NewTestFedoraCloudVM("start-execute-too-low-timeout-stop-vm").WithCloudConfig(fedoraCloudConfig).Build(),
+				Secret: testobjects.NewTestSecret(sshConnectionInfo),
+				Script: sleepScript,
+				// the VM is never started beforehand, so 2s is structurally too short for a
+				// fresh VMI to even reach Running - this fails deterministically regardless
+				// of cluster speed, unlike a "usually enough" value
+				Timeout: &metav1.Duration{
+					Duration: 2 * time.Second,
```
0xFelix, line 467: Can it be set to `0`?

ksimon1: No, because it means run indefinitely - https://github.com/kubevirt/kubevirt-tekton-tasks/blob/main/modules/execute-in-vm/pkg/utils/utils.go#L10

test/execute_and_cleanup_vm_test.go
ksimon1: fixed

0xFelix (commented):
/approve

PR #938, opened 2026-07-24
feat(disk-uploader): delete VirtualMachineExport after use

### LINKED ISSUE #909: Clean up VirtualMachineExport objects immediately after use in disk-uploader
**Is your feature request related to a problem? Please describe**:
disk-uploader creates a VirtualMachineExport object (and its associated virt-export-* pod) to download a VM's disk image, then never deletes it - cleanup is left to the VirtualMachineExport after a 2hr TTL deadline. I discovered this with recent testing on my local cluster when re-running the disk-uploader TaskRun against the same PVC while the previous export's pod is still alive, this causes the new run to sit blocked in Pending for however much of the 2 hours remains.

**Describe the solution you'd like**:
Have disk-uploader explicitly delete the VirtualMachineExport it creates once the disk has been downloaded and the export is no longer needed, rather than depending solely on the TTL deadline to expire.

**Additional context**:
This is likely most visible specifically on single-node/LVM-backed storage, or any RWO access mode storage. RWX-capable storage this will likely just be a lingering, unused pod until the TTL deadline expires, rather than an issue that will block a subsequent run.

ozzytc:
disk-uploader creates a VirtualMachineExport object (and its associated virt-export-* pod) to download a VM's disk image, then never deletes it - cleanup is left to the VirtualMachineExport after a 2hr TTL deadline. This PR explicitly deletes the VirtualMachineExport object once the disk download completes, rather than relying solely on it to timeout.

Fixes #909 

**Release note**:

```release-note
feat(disk-uploader): delete VirtualMachineExport after use
```

### DIFF
--- a/modules/disk-uploader/pkg/vmexport/vmexport.go
+++ b/modules/disk-uploader/pkg/vmexport/vmexport.go
@@ -51,6 +51,10 @@ func CreateVirtualMachineExport(virtClient kubecli.KubevirtClient, exportSourceK
 	return virtClient.VirtualMachineExport(exportSourceNamespace).Create(context.Background(), v1VmExport, metav1.CreateOptions{})
 }
 
+func DeleteVirtualMachineExport(virtClient kubecli.KubevirtClient, namespace, name string) error {
+	return virtClient.VirtualMachineExport(namespace).Delete(context.Background(), name, metav1.DeleteOptions{})
+}
+
 func WaitUntilVirtualMachineExportReady(client kubecli.KubevirtClient, namespace, name string) error {
 	pollInterval := 60 * time.Second
 	pollTimeout := 3600 * time.Second
--- a/modules/disk-uploader/pkg/vmexport/vmexport.go
+++ b/modules/disk-uploader/pkg/vmexport/vmexport.go
@@ -52,7 +52,11 @@ func CreateVirtualMachineExport(virtClient kubecli.KubevirtClient, exportSourceK
 }
 
 func DeleteVirtualMachineExport(virtClient kubecli.KubevirtClient, namespace, name string) error {
-	return virtClient.VirtualMachineExport(namespace).Delete(context.Background(), name, metav1.DeleteOptions{})
+	err := virtClient.VirtualMachineExport(namespace).Delete(context.Background(), name, metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	return nil
 }
 
 func WaitUntilVirtualMachineExportReady(client kubecli.KubevirtClient, namespace, name string) error {
--- a/cmd/disk-uploader/main.go
+++ b/cmd/disk-uploader/main.go
@@ -86,6 +86,12 @@ func run(opts parse.CLIOptions, k8sClient kubernetes.Interface, virtClient kubec
 		return "", err
 	}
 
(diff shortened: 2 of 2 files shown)


ozzytc:
hi @codingben, please review when you can, thanks

modules/disk-uploader/pkg/vmexport/vmexport.go
```diff
@@ -51,6 +51,10 @@ func CreateVirtualMachineExport(virtClient kubecli.KubevirtClient, exportSourceK
 	return virtClient.VirtualMachineExport(exportSourceNamespace).Create(context.Background(), v1VmExport, metav1.CreateOptions{})
 }
 
+func DeleteVirtualMachineExport(virtClient kubecli.KubevirtClient, namespace, name string) error {
+	return virtClient.VirtualMachineExport(namespace).Delete(context.Background(), name, metav1.DeleteOptions{})
+}
```
0xFelix, line 60: IMO it would be fine to inline this

ksimon1: I agree

codingben (commented):
I'm not sure about this, but I won't block this PR. The disk-uploader's pod is the owner of this object, when the pod is deleted, all owned objects (including VMExport) will be deleted.

The cleanest solution here is to delete the disk-uploader pod, that will delete the VMExport object and other objects, why you can't delete it? How do you deploy it?

ozzytc:
@codingben that's fair, and deleting the disk-uploader pod manually is what i've been doing, however it is a manual step and it does mean I lose the local pod logs, and since the VMExport is not needed after the download completes, that seemed like the right point to free it rather than waiting for the full lifetime.

Happy to close this if you still don't see value though.

codingben (approved):
/lgtm

codingben:
> @codingben that's fair, and deleting the disk-uploader pod manually is what i've been doing, however it is a manual step and it does mean I lose the local pod logs, and since the VMExport is not needed after the download completes, that seemed like the right point to free it rather than waiting for the full lifetime.
> 
> Happy to close this if you still don't see value though.

I'm okay with this because I see that the [CLI](https://kubevirt.io/user-guide/storage/export_api/#ttl-time-to-live) supports it, there is a flag to delete VMExport after download is finished (but it's optional).

ozzytc:
thanks @codingben, re the CLI, I took a closer look at how it handles this, and I'm thinking it might be better to move the delete into a defer right after the vmexport creation, like the pattern in virtctl, so if a failure happens before the delete point the delete will still occur. What are your thoughts on this?

codingben:
> thanks @codingben, re the CLI, I took a closer look at how it handles this, and I'm thinking it might be better to move the delete into a defer right after the vmexport creation, like the pattern in virtctl, so if a failure happens before the delete point the delete will still occur. What are your thoughts on this?

In that case the error will tell us why - and deleting the VMExport itself will not allow us to debug it. This CLI is not spawning any pod but executed from the developer's environment, so you get the logs and can track it from your own machine.

modules/disk-uploader/pkg/vmexport/vmexport.go
ozzytc: Thanks for the suggestion, a separate automated review comment recommended adding IsNotFound handling here to avoid logging a misleading failure warning when the object is already gone, so I've kept this as its own function as it's no longer just a passthrough.

codingben (approved):
/lgtm

PR #965, opened 2026-08-11
fix: replace dockerhub fedora image with quay image, fix race condition in execute-in-vm task

ksimon1:
fix: replace dockerhub fedora image with quay image

It looks like dockerhub reduced again pull limit.
Replace the dockerhub image with quay image

fix(execute-in-vm): report timeout when budget expires before script runs

The shared timeout budget across EnsureVMRunning, SetupConnection, and
RemoteExecute silently skipped phases when the budget was exhausted,
causing the task to succeed instead of failing with a timeout error.

Replace the callback-based WithTimeout with a straightforward deadline
check between each phase, and report a descriptive error when the
budget runs out.

Version updates will be done via kubevirt CI in follow up PR
Fixes #

**Release note**:

```release-note
NONE
```


modules/execute-in-vm/pkg/utils/utils.go
```diff
@@ -1,19 +1 @@
 package utils
```
akrejcir: This file can be deleted too.

ksimon1: removed

codingben (changes requested):
Can you update the PR title and description to reflect the changes of another commit here?

codingben (approved):
/lgtm

Thanks Karel, looks good from my side!

